PR #30, opened 2019-07-22
HashMap, Heap and HashSet

LucunJi:
list in scarpet is based upon java's ArrayList. it was too inefficient when I try A* pathfinding in scarpet, so several utility programs are made for better time cost. some flaws may exist and depreciated method element() is used. the programs are also published [here](https://github.com/LucunJi/scarpet-stuff). planning to make hashheap and linkedlist, and functions for vectors and matrix are also planned.

some test data: HashMap can be 6.75 times faster than list in performing 1M contains() operation; and Heap is over 100 times faster to run 10k add-sort operations.

maybe scarpet can have more types like dict(), linked() and tree() so processing data can be faster sometimes

### DIFF
--- a/programs/utilities/hashmap.sc
+++ b/programs/utilities/hashmap.sc
@@ -1,0 +1,104 @@
+/script run 
+	'Author:	LucunJi(禄存)
+	 Reference:	https://www.cnblogs.com/skywang12345/p/3310835.html
+	 Thanks:	gnembon';
+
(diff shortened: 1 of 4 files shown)


gnembon:
Awesome work, thank you. dict is planned to be added as a crucial part, but some more thought needs to be put into definition of a hash value of any type of value, i.e. is '2' different from 2, etc. Never thought about linked list and tree, but that might be too complex for a simple language.

PR #35, opened 2019-09-10
Added implementation of Silk Spawners

PegasusEpsilon:


PegasusEpsilon:
Built angel blocks from RandomThings while I was at it.

PR #38, opened 2019-10-20
Create survival/bridge.sc

Aplet123:
If holding a golden sword in mainhand and sneaking, blocks in the offhand will automatically place to form a bridge up to 5 blocks long

### DIFF
--- a/programs/survival/bridge.sc
+++ b/programs/survival/bridge.sc
@@ -1,4 +1,5 @@
 global_bridge_item = 'golden_sword';
+global_bridge_lengths = l(2, 3, 4, 5, 6, 7, 8, 9, 10, 15, 20, 30, 50, 200);
 
 __can_bridge(entity) -> 
 (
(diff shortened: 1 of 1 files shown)


programs/survival/bridge.sc
```diff
@@ -0,0 +1,49 @@
...
+	if (entity ~ 'gamemode' == 'adventure', return(false));
+	if (!(entity ~ 'sneaking'), return(false));
+	holds_mainhand = query(entity, 'holds', 'mainhand');
+	holds_offhand = query(entity, 'holds', 'offhand');
+	if (!holds_mainhand || !holds_offhand, return(false));
+	if (!(holds_mainhand:0 ~ global_bridge_item), return(false));
+	true
+);
+
+__get_direction_vector(face) ->
```
gnembon: there is already a function that does that natively that you can use, `pos_offset`:
https://gnembon.github.io/scarpet/carpet/script/CarpetExpression.html#pos_offset
(sorry link won't work, you need to search)

programs/survival/bridge.sc
```diff
@@ -0,0 +1,49 @@
...
+	if(player ~ 'gamemode' != 'creative', __remove_offhand(player));
+	schedule(5, '__place_block', player, pos + dir, dir, num - 1)
+);
+
+__on_player_right_clicks_block(player, item_tuple, hand, block, face, hitvec) ->
+(
+	if (hand != 'mainhand', return());
+	if(__can_bridge(player), (
+		dir = __get_direction_vector(face);
+		__place_block(player, pos(block) + dir, dir, 5)
```
gnembon: Really cool and simple application idea, but why just only 5 blocks?

Also quite dangerous when you try to pillar up

Aplet123: I could raise the block limit, but I'm not sure what it should be. It could maybe even bridge until you run out of blocks. Also, I'm not quite sure how to fix pillaring up, the only idea I have is teleporting the player one block up if they're on top of the block being placed, although that might feel clunky.

gnembon: if its just for bridging out, just to make sure the offset on Y is 0, or the direction (face) is not `up`, or `down`. Also you can use inventory_find to get the block in the offhand and pull from the main inventory first - this will leave the offhand as the last resort, thus preventing replacing the stack after each 64 blocks. This may mean you could pillar up too far, but you can also enforce player sneaking to continue scheduling new events. If the player unshifts, that would stop building the bridge. Or just check the item in main hand, so you could run on it while its building. 

Throwing some ideas right there.

Aplet123: I think that bridging up and down really comes in handy, so I think for bridging up the player will be teleported a block up if the are inside the block being placed. Also, I feel like pulling from the offhand acts as a nice cap to the bridging, because holding sneak feels a little bit annoying.

PR #49, opened 2020-01-04
Reviving corals with water bottles

Nameless-123:
By: Ghoulboy (that name was taken)
This is a script with enables the making of a coral block farm with carpet mod's `renewableCorals` set to `true`

### DIFF
--- a/programs/survival/revive_coral
+++ b/programs/survival/revive_coral
@@ -1,16 +1,14 @@
 //allows player to revive dead corals with water bottles
 __on_player_right_clicks_block (player, item_tuple,hand, block,face,hitvec) -> (
 	l(item, count, nbt) = item_tuple || l('None', 0, null);
-	dead=l('dead_brain_coral_block','dead_bubble_coral_block','dead_fire_coral_block','dead_horn_coral_block','dead_tube_coral_block');
-	live=l('brain_coral_block','bubble_coral_block','fire_coral_block','horn_coral_block','tube_coral_block');
-	listpos=dead ~ block;
-	if(item=='potion'&& dead ~ block !=null,
-		print('WOrks');
-		set(pos(block),get(live,listpos));
+	position=pos(block);
+	types=l('brain','bubble','fire','horn','tube');
+	block=block-'dead_'-'_coral_block';
+	if(item=='potion'&& types ~ block != null && nbt=='{Potion:"minecraft:water"}',
+		set(position,block+'_coral_block');
 		gamemode=query(player,'gamemode');
-		if(gamemode!='creative',  //if player is in creative, will retain water bottle
-			slot=query(player,'selected_slot');
-			inventory_set(player,slot,1,'glass_bottle');
+		if(gamemode!='creative',
+			inventory_set(player,query(player,'selected_slot'),1,'glass_bottle');
 		);
 	);
 );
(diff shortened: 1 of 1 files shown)


programs/survival/revive_coral
```diff
@@ -0,0 +1,16 @@
+//allows player to revive dead corals with water bottles
+__on_player_right_clicks_block (player, item_tuple,hand, block,face,hitvec) -> (
+	l(item, count, nbt) = item_tuple || l('None', 0, null);
+	dead=l('dead_brain_coral_block','dead_bubble_coral_block','dead_fire_coral_block','dead_horn_coral_block','dead_tube_coral_block');
```
gnembon: you could just detect if blocks matches 'dead_\w+_coral_block` and remove 'dead_' from the block name when setting. A suggestion

programs/survival/revive_coral
```diff
@@ -0,0 +1,16 @@
...
+	live=l('brain_coral_block','bubble_coral_block','fire_coral_block','horn_coral_block','tube_coral_block');
+	listpos=dead ~ block;
+	if(item=='potion'&& dead ~ block !=null,
```
gnembon: is 'potion' also applying to other potions, not just water bottles?

programs/survival/revive_coral
```diff
@@ -0,0 +1,16 @@
...
+		print('WOrks');
```
gnembon: Now, we know it WOrks, you can remove it :)

programs/survival/revive_coral
Nameless-123: I have no clue how to do, could you please explain further?

programs/survival/revive_coral
Nameless-123: o crap I forgot thanks

Nameless-123:
Sorry, this is like my first function scarpet script, and im new to programming, so I made a ton of stupid errors

programs/survival/revive_coral
Nameless-123: yeah, true

Nameless-123:
Also, why does the '_on_player_right_clicks_block' method need all the parameters for it to work? I mean, I only need the 'player', 'item_tuple' and 'block' parameters, so why do I need the rest?

programs/survival/revive_coral
Nameless-123: never mind, it took 15 mins of testing to figure out

gnembon:
the rest is there in case you need it. if you don't need it don't use it. Some people may need those that's why.

Nameless-123:
I get that, but I don't see why it doesn't work if I don't use params, unless its easier to code that way, in which case that's all the explanation I need

programs/survival/revive_coral
```diff
@@ -0,0 +1,14 @@
+//allows player to revive dead corals with water bottles
+__on_player_right_clicks_block (player, item_tuple,hand, block,face,hitvec) -> (
+	l(item, count, nbt) = item_tuple || l('None', 0, null);
+	position=pos(block);
+	types=l('brain','bubble','fire','horn','tube');
+	block=block-'dead_'-'_coral_block';
+	if(item=='potion'&& types ~ block != null && nbt=='{Potion:"minecraft:water"}',
```
gnembon: that would work if that's all what's in the nbt tag. You could make it more robust by using `nbt:'Potion' == 'minecraft:water'`

gnembon:
when the function has incorrect number of arguments, it is not plugged in to handle that event, since scarpet makes sure the function is correct and has correct argument count.

programs/survival/revive_coral
Nameless-123: thx I will use

PR #58, opened 2020-01-06
Easier renewable sponge farm in 1.14+

Nameless-123:
There is no rng manipulation as far as I know, so sponge farms are near impossible to make useful
By:Ghoulboy

### DIFF
--- a/programs/survival/easier_renewable_sponge.sc
+++ b/programs/survival/easier_renewable_sponge.sc
@@ -1,5 +1,5 @@
 __on_player_attacks_entity(player,entity)->(
-	if(entity=='Pufferfish'||'Tropical fish'||'Salmon'||'Cod', 
+	if(has(global_fishez,str(entity))&& entity ~ 'health'==0,
 		schedule(0,'__add_dead_fish',entity);//if not the fish's health isn't 0 and I cant check if its dead or not
 	);
 );
@@ -7,12 +7,14 @@ global_puff=0;
 global_trop=0;
 global_salmon=0;
 global_cod=0;
+global_fishez = m('Pufferfish', 'Tropical Fish', 'Salmon','Cod');
+
 
 __add_dead_fish(entity)->(
-	if (entity=='Pufferfish'&& query(entity,'health')==0,
+	if (entity=='Pufferfish',
 		global_puff+=1;
 	);
-	if (entity=='Tropical Fish'&& query(entity,'health')==0,
+	if (entity=='Tropical Fish',
 		global_trop+=1;
 	);
 	if (entity=='Salmon'&& query(entity,'health')==0,
--- a/programs/survival/easier_renewable_sponge.sc
+++ b/programs/survival/easier_renewable_sponge.sc
@@ -1,0 +1,29 @@
+__on_player_attacks_entity(player,entity)->(
+	schedule(0,'__add_dead_fish',entity)//if not the fish's health isn't 0 and I cant check if its dead or not
+);
+global_puff=0;
+global_trop=0;
+global_salmon=0;
+global_cod=0;
+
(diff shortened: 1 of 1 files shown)


gnembon:
I assume you would want that instead of carpet rule for renewableSponges and without messing up with datapacks, which is fine.

programs/survival/easier_renewable_sponge.sc
```diff
@@ -0,0 +1,29 @@
+__on_player_attacks_entity(player,entity)->(
+	schedule(0,'__add_dead_fish',entity)//if not the fish's health isn't 0 and I cant check if its dead or not
+);
+global_puff=0;
+global_trop=0;
+global_salmon=0;
+global_cod=0;
+
+__add_dead_fish(entity)->(
```
gnembon: global_sponge_fishez = m('Pufferfish', 'Tropical Fish', ...);
..
if (has(global_sponge_fishez, str(entity)) && entity ~ 'health' == 0, ...)

programs/survival/easier_renewable_sponge.sc
```diff
@@ -0,0 +1,29 @@
+__on_player_attacks_entity(player,entity)->(
+	schedule(0,'__add_dead_fish',entity)//if not the fish's health isn't 0 and I cant check if its dead or not
```
gnembon: I woudn't schedule call for every player kill - also check the entity type here

programs/survival/easier_renewable_sponge.sc
```diff
@@ -0,0 +1,29 @@
...
+	if (entity=='Tropical Fish'&& query(entity,'health')==0,
+		global_trop+=1;
+	);
+	if (entity=='Salmon'&& query(entity,'health')==0,
+		global_salmon+=1;
+	);
+	if (entity=='Cod'&& query(entity,'health')==0,
+		global_cod+=1;
+	);
+	if(global_puff>0 && global_trop >0 && global_salmon>0 && global_cod>0 &&rand(50)==0,
```
gnembon: I assume you will start getting spongs when you kill at least one of each?
if you would keep player stats in a map, then  you could use all(...) function. But that would reset with every restart - I assume that's intended

programs/survival/easier_renewable_sponge.sc
Nameless-123: ok, i'll fix that

programs/survival/easier_renewable_sponge.sc
Nameless-123: ok yeah, I have no clue clue how to use any form of map. I always end up with northe pointing wst and south pointing east

programs/survival/easier_renewable_sponge.sc
Nameless-123: yeah, i totally understand...

programs/survival/easier_renewable_sponge.sc
gnembon: in this case map or list, doesn't matter that much, but either would look much better than several if's next to each other.

Nameless-123: ok, yeah, i have no clue how to do this wth lists either

programs/survival/easier_renewable_sponge.sc
Nameless-123: I dont know how to do the rest so ill just do it for the more obvious section

gnembon:
hmm, seems like I forgot to mention - please replace run(`/summon, with spawn(...) and we can merge like that.

Nameless-123:
> hmm, seems like I forgot to mention - please replace run(`/summon, with spawn(...) and we can merge like that.

It never seems to work when I try, so I can't seem to do it

PR #60, opened 2020-01-07
Adding one survival app, and adding readmes to most of the apps folders

indoorjetpacks:
Added signs_for_water_streams - skips the message entry prompt when a bucket is in the offhand
Added readmes to most of the apps folders, not entirely sure how accurate they are, just tried to open up the apps and generalize them a bit into a category

### DIFF
--- a/programs/fundamentals/README.md
+++ b/programs/fundamentals/README.md
@@ -1,1 +1,1 @@
-Useful references for basic math operations, and programs dealing with hashes and hex values
+Programs for calculating math operations, and examples for min/max interation, heaps and hashes
--- a/programs/fundamentals/README.md
+++ b/programs/fundamentals/README.md
@@ -1,0 +1,1 @@
+Useful references for basic math operations, and programs dealing with hashes and hex values
--- a/programs/simulations/README.md
+++ b/programs/simulations/README.md
@@ -1,1 +1,1 @@
-Examples for using generators, and setting large numbers of blocks
+Examples for setting pallettes of blocks, setting sizes to change large areas of blocks based on input conditions (either player or mathmatically defined ones)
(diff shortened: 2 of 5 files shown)


gnembon:
Thanks for all the readmes - its easier to edit later than add.
The skipping of sign dialog is really useful, but would it be better if it would only check if player is sneaking to skip it?
you can replace consecutive it statements with map lookup.
Also, the `:` operator can accept nulls as left operands, so most of that code could be achieved with 
```
global_item_to_block = m( l('birch_sign', 'birch_wall_sign'), ...);
__on_player_right_clicks_block(player, item_tuple, hand, block, face, hitvec) ->
(
   if(has(global_item_to_block(item_tuple:0)) && player ~ 'sneaking', place_block(global_item_to_block:(item_tuple:0), ....
```

indoorjetpacks:
Yeah that makes sense about a sneaking check, keeps it consistent with how the vanilla game lets you sneak to ignore other GUI's like hoppers, droppers, etc. 
Looking through the code and trying to understand it - the map and list bit makes sense I think, it's mapping a list of each of the item ID names to the wall sign ID names. 
Ah so the `:` operator will get a null value from the mapping then since non-listed values will return null from a map, which makes the if statement not return true for that. Okay cool thanks. 

Is `place_block()` pseudo code? I was trying to use `place_item()` to avoid having to check for air blocks first and using `set()`, and since I think it should be considered a player placement instead of the code 'setting' it - but wasn't able to get 'place_item()' to work for some reason. No error messages or anything though, which was strange.

gnembon:
place_item works, just places standard standing sign and not wall_sign. I could try to debug it. There is nothing wrong with using `set()`. If you want I can merge this PR and you can keep working on it,ok?

indoorjetpacks:
Sounds good! Eh if using `set()` is fine then I'd say don't worry about debugging the place_item for it until it comes up again. But yeah, you can go ahead and merge and I'll keep working on it.

PR #69, opened 2020-01-28
Improved README.md from the garbled mess that it was before

Ghoulboy78:
I added proper styling for app names, creator names, then a creator list for those who wanted to know who something.
Also, I added clear instructions at the top.


Ghoulboy78:
Accident sorry

PR #79, opened 2020-02-26
Added a load of programs

Ghoulboy78:
added shear_corals.sc Readme and the script itself(I hope)


Ghoulboy78:
Please ignore that last bit, I was just messing around in my fork

PR #90, opened 2020-03-23
Survival lighter.sc - It works like the previous, but in a 128 box around player

Ghoulboy78:
and it consumes torches in survival and gives weird error messages if you run out of torches, so yay
and you can add rand if you want, but i found it kinda pointless, the user can chooze

### DIFF
--- a/programs/survival/survival_lighter.sc
+++ b/programs/survival/survival_lighter.sc
@@ -6,7 +6,7 @@ __on_player_uses_item(player, item_tuple, hand)->(
 		scan(cx,cy,cz,128,128,128,
 			if(!inventory_find(player,'torch'),return(print('You ran out of torches')));//it gives weird error messages in surival for some reason, but works otherwise.
 			lpos=pos(_);
-			if((block(lpos)=='air'||block(lpos)=='cave_air')&&block_light(lpos)<8&&sky_light(lpos)<8&&solid(pos_offset(lpos, 'down')),//&&!rand(10), optional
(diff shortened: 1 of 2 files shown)


Ghoulboy78:
however, it looks less cool than the previous, ill say that much

programs/survival/survival_lighter.sc
```diff
@@ -0,0 +1,17 @@
+__on_player_uses_item(player, item_tuple, hand)->(
+	l(item,count,nbt)=item_tuple;
+	if(hand=='mainhand'&&item == 'torch'&&nbt==null,
+		print('Started placing torches ... ');//You can turn off feedback
+		l(cx,cy,cz)=pos(player);
+		scan(cx,cy,cz,128,128,128,
+			if(!inventory_find(player,'torch'),return(print('You ran out of torches')));//it gives weird error messages in surival for some reason, but works otherwise.
+			lpos=pos(_);
+			if((block(lpos)=='air'||block(lpos)=='cave_air')&&light(lpos)<8&&solid(pos_offset(lpos, 'down')),//&&!rand(10), optional
```
gnembon: air(pos) will check for you 'air', 'cave_air' as well as 'void_air' in one expression

Ghoulboy78: done

gnembon:
Why don't you just add survival check for item usage to the previous one? - its a cool demonstration no how to control falling block entities in scarpet plus it looks cool - the only thing its missing is checks for survival to remove the item (making sure the torch stack in the hand is the last to be used).

Ghoulboy78:
why should the item in the mainhand be the last to be used? and ok, perhaps i will try und find a way

Ghoulboy78:
you shouldn't have man, I could have done it! I mean, I just had a little extra work this week.

PR #91, opened 2020-03-23
It works like the previous, but in a 128 box around player

Ghoulboy78:
and it consumes torches in survival and gives weird error messages if you run out of torches, so yay
and you can add rand if you want, but i found it kinda pointless, the user can chooze


Ghoulboy78:
accident soz

PR #96, opened 2020-04-25
i dont even know if this is necessary.

Ghoulboy78:
Heck, you could just move the scarpet apps to the fabric-carpet repo to save yourself some work. Just put the programs directory next to the docs, or even better, just rename the docs folder to scarpet, and then rename the internal one from scarpet to docs, and next to it put the programs directory from here, saving yourself a lot of work going back and forth.
You should just make the readme a wiki page. In fact, lemme do that


Ghoulboy78:
gnembon, I will make it my task to keep the fabric-carpet wiki page on scarpet up to date, don't you worry!

PR #97, opened 2020-05-06
Added lifetimes.sc

Ghoulboy78:
to measure mob lifetimes in a nice and user-friendly way

### DIFF
--- a/programs/utilities/lifetimes
+++ b/programs/utilities/lifetimes
@@ -1,0 +1,72 @@
+__command()->(
+    print(
+        'Available commands:\n'+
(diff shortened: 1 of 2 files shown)


programs/utilities/lifetimes
```diff
@@ -0,0 +1,72 @@
+__command()->(
```
gnembon, line 1:
```suggestion
// by Ghoulboy78

__command()->(
```

Ghoulboy78: oh yeh forgot

PR #99, opened 2020-05-24
Create template_script.sc

Ghoulboy78:
This is a template, feel free to use it or not.
I personally found it useful so I can understand code flow better, and so it's not a jamble to read.


gnembon:
honestly not really needed. It all depends on the application what you need.

PR #103, opened 2020-06-01
Coordinate Exporter

aria1th:
coordinate exporter app, which allows you to get coordinates of structures in defined range. 

It does not load or generate map, so it could be very 'light'. It is slower than external amidst, but it is more precise than amidst, especially when you tries to find end cities, or end ships. Since this app does not offer any copy-paste or save options by itself, you need clients that provides logging to copy and paste, to use in other external apps.

current implemented options
- amidst to check all structures
- warped forest fortress finder
- end ship finder (finds end city with ship exclusively)
- nether structure finder

### DIFF
--- a/programs/survival/cam.sc
+++ b/programs/survival/cam.sc
@@ -1,173 +1,68 @@
-// safe and fair spectator camera app by gnembon
-// saves player's position, motion, angles, effects and restores them on landing
-// saves player configs between saves
-// places spectating players without config in a safe places
-// and adds timeout for the player so they cannot switch instantly
-// change setting below to change that behaviour (0 to disable timeout)
-global_survival_timeout = 3;
-
-// additional player checks are here. comment lines that you think are not needed
-__assert_player_can_cam_out(player) ->
-(
-   if(!query(player,'nbt','OnGround'), exit('You must be on firm ground.'));
-   if(query(player,'nbt','Air') < 300, exit('You must be in air, not suffocating nor in liquids.'));
-   if(player ~ 'is_burning', exit('You must not be on fire.'));
-   null
+__command()->(
+    print('This camera acts the same as /c, except it will tp you back to the start location, and remember previous enchantments!');
+    print('    /cam c: Will put you into spectator mode and give you night vision and conduit power to see better');
+    print('    /cam s: Will put you back into survival, give you back any conduit power and night vision effects you had before');
+    print('    /cam tpback true/false: If true, will tp you back to the start location, if not will not');
+    print('    /cam checkvalid true/false: If true, will enable some checks to be made for changing into spectator, like drowning or burning');
+    print('    /cam checkspecific check, true/false: Will toggle those checks on or off for turning into spectator');
+    print('    /cam checks: Will print all available checks, along with their default and actual values');
 );
 
-// none of your business below
+global_prevnv=l(0,0);//Night vision duration, Night vision amplification
+global_prevcp=l(0,0);//Conduit power duration, Conduit power amplification
+global_tpback=true;
+global_checkvalid=true;
+global_checkspecific=m(
+    l('falling',true),
+    l('suffocating',true),
+    l('burning',true)
+);//you can add more checks in the future
 
-
-
-
-__command() ->
-(
-   p = player();
-   current_gamemode = p~'gamemode';
-   if ( current_gamemode == 'spectator',config = __get_player_stored_takeoff_params(p~'name');
-      if (config&&(p~'dimension'==config:'dimension'),
-         __restore_player_params(p, config);
-         __remove_player_config(p~'name');
-         __remove_camera_effects(p);
-      ,
-         if (__survival_defaults(p), 
-	if(config = __get_player_stored_takeoff_params(p~'name'), __restore_player_params_irregular(p, config);__remove_player_config(p~'name'));
-            __remove_camera_effects(p)
-         );
-      );
-   , (current_gamemode == 'survival'||current_gamemode == 'creative') && !global_is_in_switching, // else if survival - switch to spectator
-      __assert_player_can_cam_out(p);
-      if (global_survival_timeout > 0,
-         global_is_in_switching = true;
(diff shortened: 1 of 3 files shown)


Ghoulboy78:
What is the point of this app?

aria1th:
Exporting coordinates of structures (or biomes) without direct access(or locate function), since amidst's  coordinates and exported coordinates were incorrect.

Ghoulboy78:
couldn't you just do `run('locate EndCIty')` and record the result in a variable? I mean, Idk if it's possible, but is it?

aria1th:
I actually tried it but I had to get appropriate distance to get it, too short distances between execution points would lead to slower result than structure_eligibility, and too far distances would miss some end cities which are close to each other. Especially, I could not find end ship with locate command.

gnembon:
yeah, good point - please add some header / description up top so someone who is curious what this does can read from it.

Ghoulboy78:
can you do that in a separate pr? this makes it more organized

aria1th:
Um I don't know how to separate it... but i'll try. I'm just jumped into learning it

Ghoulboy78:
oh ok. just undo those changes, then wait for this one to be accepted, then do those ones. Or ask someone else to do it

Ghoulboy78:
wait a min, your edit was my original cam app!

aria1th:
nooo it's messing up... how could i undo it

Ghoulboy78:
it's already done

Ghoulboy78:
like un-done

Ghoulboy78:
what I'm saying is that the edit that you were doing is the original cam app which I made. how come?

aria1th:
I didn't realized that; I was just editing repository and... it happened

Ghoulboy78:
oh nice

aria1th:
oh my it's messing up

programs/survival/cam.sc
```diff
@@ -1,68 +1,173 @@
...
+      scan(x, _, z, 32, 0, 32,
+         if( air(_) && air(pos_offset(_, 'up')) && solid(pos_offset(_, 'down')),
+            modify(player, 'pos', pos(_)+l(0.5,0.7,0.5));
+            return(true);
+         )
+      )
+   );
+   print(format('rb Cannot find a safe spot to land within 32 blocks away'));
+   false;
+);
```
Ghoulboy78, line 173: change to 
__command()->(
    print('This camera acts the same as /c, except it will tp you back to the start location, and remember previous enchantments!');
    print('    /cam c: Will put you into spectator mode and give you night vision and conduit power to see better');
    print('    /cam s: Will put you back into survival, give you back any conduit power and night vision effects you had before');
    print('    /cam tpback true/false: If true, will tp you back to the start location, if not will not');
    print('    /cam checkvalid true/false: If true, will enable some checks to be made for changing into spectator, like drowning or burning');
    print('    /cam checkspecific check, true/false: Will toggle those checks on or off for turning into spectator');
    print('    /cam checks: Will print all available checks, along with their default and actual values');
);

global_prevnv=l(0,0);//Night vision duration, Night vision amplification
global_prevcp=l(0,0);//Conduit power duration, Conduit power amplification
global_tpback=true;
global_checkvalid=true;
global_checkspecific=m(
    l('falling',true),
    l('suffocating',true),
    l('burning',true)
);//you can add more checks in the future

c()->(
    player=player();
    valid=__valid(player);//will return null if you can go into spectator mode
    if(valid,return('You cannot go into spectator while '+valid));
    run('gamemode spectator '+player);
    global_prevnv=query(player,'effect','night_vision');
    global_prevcp=query(player,'effect','conduit_power');
    modify(player, 'effect', 'night_vision', 0, 0, false, false);//remove pre-existing effects
    modify(player, 'effect', 'conduit_power', 0, 0, false, false);//cos otherwise its buggy
    modify(player, 'effect', 'night_vision', 1000000, 255, false, false);
    modify(player, 'effect', 'conduit_power', 1000000, 255, false, false);
    global_pos=pos(player);
    return(null)
);

s()->(
    player=player();
    if(!query(player,'gamemode_id'),return('Can\'t use command in survival!'));//cos survvial is 0, so you cant use to to teleport back to a spot
    run('gamemode survival '+player);
    modify(player, 'effect', 'night_vision', get(global_prevnv,1), get(global_prevnv,1), true, true);
    modify(player, 'effect', 'conduit_power', get(global_prevcp,1), get(global_prevcp,1), true, true);
    if(global_tpback,modify(player,'pos',global_pos));//tps you back to start
    return(null)
);

tpback(tpback)->global_tpback=tpback;
checkvalid(checkvalid)->global_checkvalid=checkvalid;
checkspecific(check,val)->global_checkspecific:check:1=val;//I think its the correct syntax

checks()->(
    player=player();
    print('All currently available checks to turn into spectator mode');
    print('- falling '+global_checkspecific:0);
    print('- suffocating '+global_checkspecific:1);
    print('- burning '+global_checkspecific:2);
);

__valid(player)->(
    cause = null
    if(!global_checkvalid, return cause);// if checking for validity is off, return null
    if(global_checkspecific:0&& !(query(player,'nbt','OnGround')||player~'jumping'),cause='falling');//falling, but not jumping
    if(global_checkspecific:1&& query(player,'nbt','Air')<300,cause='suffocating');//suffocating, and also bobbing up and down in water, eg. waterlogged trapdoor
    if(global_checkspecific:2&& player~'burning',Scause='burning');//burning, and any others?
    return cause
);

//Conditions for cheating (copy pastad code from Minerva Datapack where I originally devised these checks)
//execute unless data entity @s {Dimension:0} run scoreboard players set @s valid 0 <- dunno about this one

Ghoulboy78: I think that will fix it

Ghoulboy78:
i think you may have to delete your fork, then refork and do this again, cos this is a mess.

aria1th:
May I ask you to open pull request instead of me? I'm having trouble with pushing my own branch. Then I can close this pull request

Ghoulboy78:
sure!

PR #105, opened 2020-06-04
Changed auto_lighter.sc to make it so you can't cheat it by going into spectator

Ghoulboy78:
Cos then if you went into spec afer starting, it would not make u lose any torches. Now that val is set before you start.
And you can't cheat it by starting in spectator either. U simply can't start it in spectator, tho u can continue it in spec

### DIFF
--- a/programs/survival/auto_lighter.sc
+++ b/programs/survival/auto_lighter.sc
@@ -12,11 +12,12 @@ __on_player_uses_item(player, item, hand) ->
 		ench = item:2:'Enchantments[]';
 		global_spread_love = 0;
 		delete(item:2:'Enchantments');
-		if (!ench,
+		if (!ench && player~'gamemode_id'!=3,
 			global_spread_love = 1;
 			if (ench==null, item:2 = nbt('{}'));
 			put(item:2:'Enchantments','[]');
 			put(item:2:'Enchantments', '{lvl:1s,id:"minecraft:protection"}', 0);
+		    	global_survival=!(player~'gamemode_id' % 2);
 			schedule(0, 'spread_torches', player);
 		);
 		inventory_set(player, player~'selected_slot', item:1, item:0, item:2);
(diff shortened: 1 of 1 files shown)


programs/survival/auto_lighter.sc
```diff
@@ -12,11 +12,12 @@ __on_player_uses_item(player, item, hand) ->
...
 		global_spread_love = 0;
 		delete(item:2:'Enchantments');
-		if (!ench,
+		if (!ench && player~'gamemode_id'!=3,
 			global_spread_love = 1;
 			if (ench==null, item:2 = nbt('{}'));
 			put(item:2:'Enchantments','[]');
 			put(item:2:'Enchantments', '{lvl:1s,id:"minecraft:protection"}', 0);
+		    	global_survival=!(player~'gamemode_id' % 2);
 			schedule(0, 'spread_torches', player);
```
gnembon, line 21: man, why globals, just add initial gamemode as an arguement to spread_torches.

gnembon, line 21:
```suggestion
			schedule(0, 'spread_torches', player, player~'gamemode_id');
```

programs/survival/auto_lighter.sc
```diff
@@ -27,10 +28,12 @@ __distance_sq(vec1, vec2) -> reduce(vec1 - vec2, _a + _*_, 0);
 
 global_effect_radius = 128;
 
+global_survival=false;
+
 spread_torches(player) ->
```
gnembon, line 33:
```suggestion
spread_torches(player, initial_gamemode) ->
```

programs/survival/auto_lighter.sc
```diff
@@ -27,10 +28,12 @@ __distance_sq(vec1, vec2) -> reduce(vec1 - vec2, _a + _*_, 0);
...
 (
 	if (global_spread_love && player~'holds':0 == 'torch',
```
gnembon, line 35:
```suggestion
	if (global_spread_love && player~'holds':0 == 'torch' && player ~'gamemode_id' == initial_gamemode,
```

gnembon: cannot add suggestions to the code below, just add initial_gamemode to the schedules for spread_torches

programs/survival/auto_lighter.sc
Ghoulboy78: ok, man, why u bi h8'in on globalization? It's not like it caused a global pandemic or anything!

Ghoulboy78:
Wait, but how come you can't edit it?
image

PR #106, opened 2020-06-04
Made it so if you are in an invalid location at time of switch, it will not put you in spec mode

Ghoulboy78:
Tho I can't seem to make it give a message


gnembon:
well these checks only apply to player in survival, so not spectators. Btw its already checked for validity when typing /cam

Ghoulboy78:
no, cos during the cooldown you can get into invalid location, and it still makes u into cam mode

PR #108, opened 2020-06-08
Added configurations to loaded chunks display

Ghoulboy78:
Modification of position and refresh rate.


gnembon:
should I maybe remove that file entirely due to chunk display being built-in carpet using client side rendering?
Its so much more practical comparing to this - block based solution.

Ghoulboy78:
true

PR #126, opened 2020-08-11
Opened a new category

Firigion:
Here people should drop their creative building tools made with scarpet, so they are clearly separate from utilities.


Firigion:
Added a few apps that I think fit the category and are polished enough to be of public use.

PR #128, opened 2020-08-14
Add Right Click Harvest

SirGavith:
I made right click harvest, figured I'd add it to the app store


SirGavith:
I'm going to wait until later versions of the script to do this pr

PR #131, opened 2020-09-17
Added vacuum.sc

rv3r:
Added vacuum scarpet program
Added description to list of programs
Added name to contributors

### DIFF
--- a/programs/survival/vacuum.sc
+++ b/programs/survival/vacuum.sc
@@ -1,3 +1,5 @@
+//first shown on Xisumavoid's YouTube channel
+//https://youtu.be/FMu8T8KriQY
 __config() ->
 (
 	return(m(l('scope','player'),l('stay_loaded',true)))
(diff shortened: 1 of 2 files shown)


programs/survival/vacuum.sc
```diff
@@ -0,0 +1,52 @@
...
+__config() ->
+(
+	return(m(l('scope','player'),l('stay_loaded',true)))
+);
+
+__move_items() ->
+(
+	loop(41,
+		//check to see if the item is a single shulker box named 'Vacuum'
+		if(inventory_get(player(), _):0 ~ 'shulker_box' && parse_nbt(parse_nbt(inventory_get(player(),_):2):'display':'Name'):'text' == 'Vacuum' && inventory_get(player(),_):1 == 1,
```
gnembon, line 12: I would suggest avoiding excessive parsing of entire shulkerbox nbt just to get its name, and also fetching the nbt only once. That should significantly cut down on compute

PR #133, opened 2020-09-19
Making it more realistic by taking into account hardness of block to apply velocity.

Ghoulboy78:
Im not sure whether or not to add a check to see if the overall speed is more than a given speed, in which case to reduce to fixed value, so as not to make it too fast.
On the other hand, the given test subject is not exactly know for realism...

### DIFF
--- a/programs/survival/holy_hand_grenades.sc
+++ b/programs/survival/holy_hand_grenades.sc
@@ -92,9 +92,9 @@ __cascade_smash(center_pos, ttl, bias, position) ->
 	if (properties_tag, put(nbttag, 'BlockState.Properties', properties_tag));
 	falling_block = spawn( 'falling_block', x+0.5, y+0.5, z+0.5, nbttag );
 	modify(falling_block,'motion',
-		(x-cx)/10+rand(0.05)+dx,
-		2*ttl/10+0.5+rand(0.2)+(y-cy)/10+dy,
-		(z-cz)/10+rand(0.05)+dz
+		((x-cx)/10+rand(0.05)+dx)/hardness(pos(x, y, z)),
+		(2*ttl/10+0.5+rand(0.2)+(y-cy)/10+dy)/hardness(pos(x, y, z)),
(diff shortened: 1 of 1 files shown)


programs/survival/holy_hand_grenades.sc
```diff
@@ -50,16 +51,14 @@ __create_bullet(player, power) ->
...
+	if (block, __cascade_smash(pos(entity), 1.2*power, direction_vec/hardness(pos(entity)), pos(block)))
 );
 
-__is_invalid_for_smashing(block) -> 
+__is_invalid_for_smashing(block) -> (
 	air(block) || block == 'lava' || block == 'water' || 
-	block == 'bedrock' || block == 'barrier' || 
-	block ~ 'command_block' || block == 'structure_block' || block == 'jigsaw';
-
+	block == 'bedrock' || item_category(block)==null;
```
gnembon, line 60: Did I miss an item category?

Ghoulboy78: Not really, just this is faster, and if more things are added in the future to this list, then this will cover them all. Also it's a good application for people who wanna see how it works.

programs/survival/holy_hand_grenades.sc
```diff
@@ -92,9 +92,9 @@ __cascade_smash(center_pos, ttl, bias, position) ->
 	if (properties_tag, put(nbttag, 'BlockState.Properties', properties_tag));
 	falling_block = spawn( 'falling_block', x+0.5, y+0.5, z+0.5, nbttag );
 	modify(falling_block,'motion',
-		(x-cx)/10+rand(0.05)+dx,
-		2*ttl/10+0.5+rand(0.2)+(y-cy)/10+dy,
-		(z-cz)/10+rand(0.05)+dz
+		((x-cx)/10+rand(0.05)+dx)/hardness(pos(x, y, z)),
```
gnembon, line 95: Do you see that the block has been set right above to air (line 86) so this has little effect here?

gnembon (approved):

PR #138, opened 2020-09-25
All true players sleeping Fixed!

Sanluli36li:
completely rewritten
#136 gnembon/fabric-carpet#425
Fix the problem that can only be reset to day 1
Now it will not check the player's sleep every tick More efficient
All behave more like Vanilla

thx gnembon's suggest

### DIFF
--- a/programs/survival/sleep_ignore_fake_players.sc
+++ b/programs/survival/sleep_ignore_fake_players.sc
@@ -1,0 +1,82 @@
+// Skip the night and no longer required fake players to sleep.
+// Different from "OnePlayerSleeping", still all real player is required on the server to cause night to pass.
+
+// Known issues: Ignore doDaylightCycle and doWeatherCycle rules, Because they can cause the player to not get up
+// I don’t know yet there is any way in Scarpet to get players up.
+
+__config() -> { 'scope' -> 'global', 'stay_loaded' -> true };
+
+global_allPlayerSleep = false;
+
+__on_tick() ->
+(
+    if(global_allPlayerSleep && __check_sleeping_enough(),
+        global_allPlayerSleep = false;
+        if(run('gamerule doDaylightCycle'),
+            l = day_time() + 24000;
+            day_time(l - l % 24000);
+        );
+        if(run('gamerule doWeatherCycle'), run('weather clear'));
+    );
+);
+
+// player login
+__on_player_connects(player) ->
+(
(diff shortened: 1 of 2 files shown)


Sanluli36li:
Now only executed in overworld, no longer check other dimensions

Sanluli36li:
Known issues: Ignore doDaylightCycle and doWeatherCycle rules, Because they can cause the player to not get up, I don’t know yet there is any way in Scarpet to get players up.

programs/survival/sleep_ignore_fake_players.sc
```diff
@@ -0,0 +1,84 @@
...
+// player wakes up
+__on_player_wakes_up(player) ->
+(
+    __update_sleeping_players();
+);
+
+// player interacts with bed
+__on_player_interacts_with_block(player, hand, block, face, hitvec) ->
+(
+    for([
```
gnembon, line 46: you can just check if the block ends with '_bed' using `block ~ '_bed$'`

programs/survival/sleep_ignore_fake_players.sc
```diff
@@ -0,0 +1,84 @@
...
+        );
+        global_allPlayerSleep = j > 0 && j >= length(entity_list('player')) - i;
+    );
+);
+
+__check_sleeping_enough() ->
+(
+    for(entity_list('player'),
+        if(_~'gamemode' == 'spectator' || _~'player_type' == 'fake' || query(_, 'nbt', 'SleepTimer') >= 100,
+            continue(),
```
gnembon, line 79: continue can just be replaced with 'null' (as a noop) and return at the end, just with `true`

programs/survival/sleep_ignore_fake_players.sc
```diff
@@ -0,0 +1,84 @@
...
+__config() -> { 'scope' -> 'global', 'stay_loaded' -> true };
+
+global_allPlayerSleep = false;
+
+__on_tick() ->
+(
+    in_dimension('overworld',
+        if(global_allPlayerSleep && __check_sleeping_enough(),
+            global_allPlayerSleep = false;
+            if(run('gamerule doDaylightCycle'),
```
gnembon, line 16: run() now returns a triple (three results) so you would need to adjust to that.

Ghoulboy78:
Add name to list of contributors

PR #139, opened 2020-09-29
Add a script to add a 33% chance to drop a players head on death

RubberDuck55:
Not Much to it, but might as well have it.


programs/survival/drop_heads.sc
```diff
@@ -0,0 +1,14 @@
+//A Balanced way to get player heads in survival.
+
+__on_player_dies(player) -> (
+  if(rand(1) >= 0.3,
```
RubberDuck55, line 4: Need To change >= to <= to make it actually 33%

gnembon (approved):

PR #142, opened 2020-10-09
Added destructive_lightning.sc

Madff386:
Once you start it any lightning that strikes, either natural, summoned or a channeling trident, a TNT and fireball will be summoned to create a small crater where the lightning strikes.

### DIFF
--- a/programs/survival/destructive_lightning.sc
+++ b/programs/survival/destructive_lightning.sc
@@ -1,0 +1,32 @@
+
+//script that makes lightning slightly more destructive
+
+__command() -> (
+    print(format('b \nDestructive Lighting'));
+    print('--------------------------------');
+    print('To enable destructive lightning:');
+    print(format(' /destructive_lightning start\n', '^ Click to start', '!/destructive_lightning start'));
+    print('To disable destructive lightning:');
+    print('Reload this script\n')
+);
+
+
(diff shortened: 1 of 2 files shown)


Ghoulboy78:
use `__on_tick()` event, not the loop.

Madff386:
Ok I updated it to use the `__on_tick()` event instead of the loop, and also added functionality to stop it without reloading.

programs/survival/destructive_lightning.sc
```diff
@@ -0,0 +1,44 @@
...
+
+__on_tick() -> (
+    if(global_started == 1,
+    e = entity_list('lightning_bolt'); //finds all lightning bolts
+        for(e,
+            look = l(0.0, -1.0, 0.0);
+            fireball_pos = (_ ~ 'pos')+look;  //finds lightning bolt position
+            if (global_last != query(_, 'uuid'),  //only summons one fireball per bolt
+                spawn('tnt', (_ ~ 'pos'));
+                game_tick(50);
```
gnembon, line 25: why do you need to run a game tick here?

Madff386: Forgot to remove it when I copied the code out of the loop and into the __on_tick event

PR #144, opened 2020-10-11
Added some BisUmTo survival and utilities scripts

BisUmTo:
**Survival Scripts Added:**
rope_ladders.sc
silk_blockstates.sc
vines_instant_mine.sc

**UtilitiesScripts Added:**
give_head.sc
gm.sc
sidebar.sc

### DIFF
--- a/programs/survival/silk_blockstates.sc
+++ b/programs/survival/silk_blockstates.sc
@@ -1,5 +1,5 @@
 ///
-// Silk Blockstates
+// Silky Blockstates
 // by BisUmTo
 // (Carpet Mod 1.4.11)
 //
@@ -7,12 +7,13 @@
 // It doesn't apply to containers and blacklisted blocks.
 ///
 
-__config() -> {'stay_loaded' -> true};
+__config() -> {'stay_loaded' -> true, 'scope' -> 'global'};
 
 global_drop_in_creative = true;
 global_create_item_whitelist = ['^spawner$', '^cake$'];
 global_preserve_block_state_blacklist = ['_bed$', '_door$', '^sticky_piston$', '^piston$', '^bee_nest$', '_banner$', '^beehive$', '^redstone_wire$'];
 global_preserve_block_data_blacklist = ['^bee_nest$', '^beehive$', '^campfire$', '^soul_campfire$', '^lectern$', '^jukebox$', '_banner$', '^player_head$', '_bed$'];
+global_need_forced_placement = ['^spawner$','_sign$'];
 
 // returns a map with all block_properties of the block and the relative value
 block_state(block) -> (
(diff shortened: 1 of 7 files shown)


programs/survival/vines_instant_mine.sc
```diff
@@ -0,0 +1,29 @@
...
+__config() -> {'stay_loaded' -> true};
+
+_vines_instant_mine(player, block) -> (
+    g = player ~ 'gamemode';
+    holds = player ~ 'holds';
+    if(g == 'spectator' || g == 'creative' || !holds || block != 'vine', return());
+    [item, count, nbt] = holds;
+    if(item!='shears',return());
+    if(g == 'adventure' && !nbt:'CanDestroy' ~ '"minecraft:vine"', return());
+    schedule(0,
```
Ghoulboy78, line 18: Correct me if this is wrong, but shouldn't it be scheduled for 1 tick delay? cos it takes 1 tick for a block to be instamined, but im not sure if this already does that.

Ghoulboy78: Im talking about the bottom line.

BisUmTo: The current code breaks the block at the end of the current gametick, so you can mine another block in the next tick (20 blocks mined each second).
I think that's correct, but I'm not 100% sure

Ghoulboy78: ok

Ghoulboy78: I can't close this.

programs/survival/silk_blockstates.sc
```diff
@@ -0,0 +1,162 @@
...
+);
+
+// returns the escaped version of the string (\->\\, d->\d)
+escape_string(string,d) -> (
+    replace(replace(string, '\\\\', '\\\\\\\\'), d, '\\\\'+d)
+);
+
+// returns true if the block is repleaceable while trying to place a block in its position
+replaceable(block) ->(
+    ['air', 'bubble_column', 'cave_air', 'crimson_roots', 'dead_bush', 'fern', 'fire', 'grass', 'large_fern', 'lava','nether_sprout','sea_grass', 'soul_fire', 'structure_void', 'tall_grass', 'vines', 'void_air', 'warped_roots', 'water'] ~ block != null || (block == 'snow' && property(block, 'layers') == 1)
```
gnembon, line 34: suggest defining these blocks outside and as a set, not list so `['foo', 'bar'] ~ block` => `global_replaceable = {'foo', 'bar'}; has(global_replaceable, str(block))`

Probably in hindsight, won't make a big difference but hey.

programs/utilities/give_head.sc
```diff
@@ -0,0 +1,69 @@
...
+    ))
+);
+
+global_utf16 = {'"'->34,'.'->46,'/'->47,'0'->48,'1'->49,'2'->50,'3'->51,'4'->52,'5'->53,'6'->54,'7'->55,'8'->56,'9'->57,':'->58,'I'->73,'K'->75,'N'->78,'S'->83,'a'->97,'b'->98,'c'->99,'d'->100,'e'->101,'f'->102,'h'->104,'i'->105,'l'->108,'m'->109,'n'->110,'p'->112,'r'->114,'s'->115,'t'->116,'u'->117,'x'->120,'{'->123,'}'->125};
+_utf16(c) -> (
+    return(global_utf16:c)
+);
+
+global_base64chars = ['A','B','C','D','E','F','G','H','I','J','K','L','M','N','O','P','Q','R','S','T','U','V','W','X','Y','Z','a','b','c','d','e','f','g','h','i','j','k','l','m','n','o','p','q','r','s','t','u','v','w','x','y','z','0','1','2','3','4','5','6','7','8','9','+','/'];
+_encode_base64(s) -> (
```
gnembon, line 36: That's some cool stuff. Should we have more ops like that (base64 encoding, etc) supported natively?

BisUmTo: That would be great! 😜

PR #145, opened 2020-10-15
Shulkerboxes: Add ability to keep one item of each stack when restocking/swapping

Senth:
Hi,
I added another feature to shulkerboxes that I found myself wanting to have. I thought I could share it since it only adds features and doesn't change any original behavior. Feel free to reject it if you want to keep the functionality simple 🙂

**The problem**
After playing around with shulkerboxes.sc on my server I had one problem. Before I installed the app I always kept one item in the shulkerbox when taking out items. This makes it easier to shift-click in items again but also refill it either manually or automatically. I also know what has been in the shulkerbox (for example my different redstone boxes) so I can't forget one component when refilling.

**Solution**
I added the ability to write `restock keep` or `swap keep` which has the same functionality as `same` but always makes sure one item of the stack is left in the shulkerbox.

**Other behaviors**
Shulkerboxes that use the `keep` keyword together with `vacuum` are now easily resupplied. If you disable the restock ability when throwing items on the ground, you can even put back everything into the shulker boxes if you throw everything on the ground.

**Tested**
Have tested and tried it with both `restock keep` and `swap keep`, and also together with `vacuum`. But no extensive testing more than playing a few hours with this feature turned on.

**Keyword**
Is the keyword `keep` good enough? I wanted to keep it short especially since it's now useful together with vacuum, but another option would be `keep1` or `keepone`. Or there might be a better word to use.

(Thanks for implementing this feature and scarpet in the first place. It makes modding and creating useful features a lot easier)

### DIFF
--- a/programs/survival/shulkerboxes.sc
+++ b/programs/survival/shulkerboxes.sc
@@ -1,6 +1,7 @@
 ///
 // vacuum and restock
 // by gnembon
+// (keep ability added by Senth)
 ///
 
 __command() -> '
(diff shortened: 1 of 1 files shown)


gnembon:
good idea.

PR #157, opened 2020-11-11
Add survival/nopickup.sc

KonaeAkira:
App that allows players to blacklist items that they do not want to pick up from the ground. Prevents unwanted items like rotten flesh from taking up precious inventory space.

### DIFF
--- a/programs/survival/nopickup.sc
+++ b/programs/survival/nopickup.sc
@@ -58,7 +58,7 @@ _count_in_inventory(item) -> (
 	while ((slot = inventory_find(player(), item:0, slot + 1)) != null, num_slots,
 		count += inventory_get(player(), slot):1;
 	);
-	return(count);
+	count;
 );
 
 // saves global_forbidden to the UUID's own config file
@@ -79,7 +79,6 @@ list() -> (
 	,
 		print('There are no forbidden items');
 	);
-	return(null);
 );
 
 add(item) -> (
@@ -91,7 +90,6 @@ add(item) -> (
 		_save_to_file();
 		print('You can no longer pick up [' + item:0 + ']');
 	);
-	return(null);
 );
 
 // sets the pickup limit for an item in the blacklist
@@ -115,7 +113,6 @@ remove(item) -> (
 	,
 		print('[' + item:0 + '] is not forbidden');
 	);
-	return(null);
 );
 
 clear() -> (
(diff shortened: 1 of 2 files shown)


programs/survival/nopickup.sc
```diff
@@ -0,0 +1,80 @@
...
+		print('There are no forbidden items');
+	);
+	return(null);
+);
+
+add(item) -> (
+	if (has(global_forbidden, str(item:0)),
+		print('[' + item:0 + '] is already forbidden');
+	,
+		put(global_forbidden, str(item:0), 0);
```
Ghoulboy78, line 57: why item:0? You can't input lists yet, only strings.

Ghoulboy78: I would suggest removing the `:0` in line 54 and 57 cos it's not a list here, it's a string.

KonaeAkira: With the new command types that scarpet supports, the input is an item (3-tuple of [name, count, nbt_data])

programs/survival/nopickup.sc
```diff
@@ -0,0 +1,127 @@
...
+
+// returns how many of "item" is in the player's inventory
+_count_in_inventory(item) -> (
+	num_slots = inventory_size(player());
+	slot = -1;
+	count = 0;
+	while ((slot = inventory_find(player(), item:0, slot + 1)) != null, num_slots,
+		count += inventory_get(player(), slot):1;
+	);
+	return(count);
```
gnembon, line 61: don't need a 'return' statement at the end, just leave `count;` on the table. It also improves performance - returns are costly

programs/survival/nopickup.sc
```diff
@@ -0,0 +1,127 @@
...
+	if (global_uuid == null, _init()); // needed in case script is reloaded when players are connected
+	if (global_forbidden != {},
+		print('List of forbidden items:');
+		for (pairs(global_forbidden),
+			print('	[' + _:0 + ', ' + _:1 + ']');
+		);
+	,
+		print('There are no forbidden items');
+	);
+	return(null);
```
gnembon, line 82: ditto with return, but this is just a command so its executed only once per command, so not a big deal.

KonaeAkira: Thank you for your input! I've made a new commit implementing the changes you suggested.

gnembon (approved):

PR #161, opened 2020-11-22
Add carried_totem.sc to survival

KingOfTheClouds:
Short script to mitigate the hand-icap of Totem of Undying (e.g. Hardcore world)

### DIFF
--- a/programs/survival/carried_totem.sc
+++ b/programs/survival/carried_totem.sc
@@ -1,7 +1,5 @@
 // Allow a Totem of Undying anywhere in the inventory to activate
 
-__config() -> ({ 'stay_loaded' -> true });
-
 __on_player_takes_damage(player, amount, source, source_entity) -> (
 	totem = 'totem_of_undying';
 	if(player~'health' <= amount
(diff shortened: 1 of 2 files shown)


programs/survival/carried_totem.sc
```diff
@@ -0,0 +1,21 @@
+// Allow a Totem of Undying anywhere in the inventory to activate
+
+__config() -> ({ 'stay_loaded' -> true });
```
gnembon, line 3: please leave config to specify for others if needed.

programs/survival/carried_totem.sc
```diff
@@ -0,0 +1,21 @@
...
+__on_player_takes_damage(player, amount, source, source_entity) -> (
+	totem = 'totem_of_undying';
+	if(player~'health' <= amount
+	&& source != 'outOfWorld'
+	&& player~'holds':0 != totem
+	&& query(player, 'holds', 'offhand'):0 != totem
+	&& inventory_remove(player, totem),
+		// borrow the offhand slot
+		if(offhand_stack = inventory_set(player, -1, 1, totem),
+			schedule(0, _(player, what, count, nbt) -> (
```
gnembon, line 14: you could technically only use `_(outer(player), outer(offhand_stack)) -> inventory_set(player, offhand_stack......)`, but that works too.

programs/survival/carried_totem.sc
KingOfTheClouds: done

programs/survival/carried_totem.sc
KingOfTheClouds: rewritten with outer

gnembon (approved):

PR #162, opened 2020-11-22
Add horse_stats.sc to survival

KingOfTheClouds:
/horse_stats prints the 3 attributes of the targeted steed

### DIFF
--- a/programs/survival/horse_stats.sc
+++ b/programs/survival/horse_stats.sc
@@ -1,7 +1,5 @@
 // Print the stats of the horse the player is looking at
 
-__config() -> ({ 'stay_loaded' -> true });
-
 __command() -> (
 	target = query(player(), 'trace', 4.5, 'entities');
 	if (target == 'Horse' || target == 'Donkey' || target == 'Mule',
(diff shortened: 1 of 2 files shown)


programs/survival/horse_stats.sc
```diff
@@ -0,0 +1,24 @@
+// Print the stats of the horse the player is looking at
+
+__config() -> ({ 'stay_loaded' -> true });
```
gnembon, line 3: suggest removing the config - it anyone wants that to load automatically they can add that.

programs/survival/horse_stats.sc
```diff
@@ -0,0 +1,24 @@
...
+
+__command() -> (
+	target = query(player(), 'trace', 4.5, 'entities');
+	if (target == 'Horse' || target == 'Donkey' || target == 'Mule',
```
gnembon, line 7: use target~'type' since named mobs will display differently here

programs/survival/horse_stats.sc
KingOfTheClouds: done

programs/survival/horse_stats.sc
KingOfTheClouds: Thanks! I overlooked query(e, 'type') in the API, though I did find the uppercase suspicious.

gnembon (approved):

PR #163, opened 2020-11-23
Added eyeremover.sc script for removing eyes of ender in peaceful

rv3r:
CCF_100 mentioned on the discord that carpet should allow for the removal of eyes of ender from end portal frames to make getting to the End possible in peaceful.

### DIFF
--- a/programs/survival/eyeremover.sc
+++ b/programs/survival/eyeremover.sc
@@ -1,13 +1,3 @@
-__config() ->
-(
-	return(
-		m(
-			l('scope','player'),
-			l('stay_loaded',true)
-		)
-	);
-);
-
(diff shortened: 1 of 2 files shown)


programs/survival/eyeremover.sc
```diff
@@ -0,0 +1,17 @@
+__config() ->
+(
+	return(
+		m(
+			l('scope','player'),
```
gnembon, line 5: I suggest not specifying scope player if its default, also stay loaded - only let people add that if they need it.
In this case we don't need

rv3r: Makes sense. That just ends up in all my programs for when they get more complex. Issue been fixed.

PR #177, opened 2021-01-17
Add new scripts

Xendergo:
Adds four scripts and a folder

I've added debug.sc, which adds a debug() function, which is like the print() function, but colors things based on their type. I've also put it in a new folder, because it didn't seem to fit into any of the others

I've also added bee_healing.sc which slowly heals bees over time, since they have a tendency to drown all the time.
I've also added compass.sc which adds a bunch of features helpful for dream style minecraft challenges
There's also shoot_fireball.sc which lets players shoot regular ghast fireballs by right clicking with fire charges. Basically the holy_hand_grenades script but more survival friendly


programs/survival/compass.sc
```diff
@@ -0,0 +1,611 @@
+__config()->{
+    'stay_loaded' -> true,
+    'commands' -> {
+        'giveCompass' -> _() -> giveCompass(entity_selector('@p'): 0),
```
gnembon, line 4: wonder why using entity selector to select a player rather than just `player()`

programs/survival/bee_healing.sc
```diff
@@ -0,0 +1,15 @@
+// Bees have an annoying habit of drowning themselves, so yeah
+__config()->{'stay_loaded' -> true, 'scope' -> 'global'};
```
gnembon, line 2: again - stay loaded is true by default - no need to specify that.

programs/survival/compass.sc
```diff
@@ -0,0 +1,611 @@
...
+        'time' -> {'type' -> 'float', 'min' -> 0.05, 'suggest' -> [0.5, 1, 10]},
+        'enabled' -> {'type' -> 'bool'},
+        'effect' -> {'type' -> 'effect'},
+        'duration' -> {'type' -> 'int', 'min' -> 0, 'max' -> 1000000, 'suggest' -> [15, 60, 1000000]},
+        'amplifier' -> {'type' -> 'int', 'min' -> 0, 'max' -> 255, 'suggest' -> [0, 1, 2, 3, 255]},
+        'dist' -> {'type' -> 'int', 'min' -> 16, 'max' -> 24000000, 'suggest' -> [8, 128, 7000, 24000000]},
+    }
+};
+
+toKeep = parse_nbt(read_file('toKeep', 'nbt'));
```
gnembon, line 45: be mindful that all of that static code is executed once per application loading and then each player gets a copy of that state when they interact with the app.

programs/survival/compass.sc
```diff
@@ -0,0 +1,611 @@
...
+        ,
+            // If there are no other players online, say so
+            print(player, format('l There are no players online'));
+        );
+    ,
+        print(player, format('l You already have the maximum amount of compasses!'));
+    );
+);
+
+item_consumed(slot) -> (
```
gnembon, line 435: just curious about it - why did you use signals to trigger item_consumed function rather than calling it directly from your app code?

programs/survival/compass.sc
```diff
@@ -0,0 +1,611 @@
...
+        if (_ == v,
+            return(true);
+        );
+    );
+
+    return(false);
+};
+
+distance(v1, v2) -> (
+    sqrt(reduce(v1, _a+(_-(v2: _i))^2, 0));
```
gnembon, line 606: suggestion for a cleaner (and possibly faster) distance function
```
sqrt(reduce(v1 - v2, _a + _*_, 0))
```

programs/survival/compass.sc
Xendergo: At the time, I didn't really trust the player() function to give the player that actually ran the command, no idea why I thought that

programs/survival/compass.sc
Xendergo: The idea is that other scripts that want to consume items can signal that event to make the persist option work with their own scripts

programs/survival/compass.sc
Xendergo: Good idea, I totally forgot you could do that

programs/survival/bee_healing.sc
Xendergo: I wrote the scripts when that wasn't case, I guess I just forgot to change it. I've removed it now tho

programs/survival/compass.sc
Xendergo: What do you mean, is that a problem?

programs/survival/compass.sc
Ghoulboy78: then shouldn't it be added to some sort of library function maybe?

programs/survival/shoot_fireball.sc
```diff
@@ -0,0 +1,27 @@
...
+        playerYaw = query(player, 'yaw')+90;
+
+        v = query(player, 'look');
+
+        modify(fireball, 'pos', playerPos+v*2); // Move the fireball in front of the player so it doesn't cover their screen for a second when it spawns
+        modify(fireball, 'nbt_merge', encode_nbt({'power' -> v*0.25})); // Set the velocity of the fireball to the direction vector
+        modify(fireball, 'nbt_merge', {'Owner' -> query(player, 'nbt', 'UUID')}); // Make the fireball count as a player kill
+        modify(fireball, 'fire', 20); // Seems to make it last longer?
+
+        if(query(player, 'gamemode') == 'survival', // Remove a fire charge from the selected slot or offhand if the player is in survival mode
```
Ghoulboy78: you can instead do `!player~'gamemode_id'%2` to capture adventure mode as well

Xendergo: Oh yeah, I totally forgot adventure mode exists. Also, what do you mean by library function? Do you mean like having a function that other scripts can import?

programs/survival/compass.sc
Ghoulboy78: I think this ought to be added to some form of library function tbh which could be built-in, which would also solve some issues in gnembon/fabric-carpet#602

Ghoulboy78: If scripts want to use this, it's easier to just copy the code. so It's better to put it in a library where it is normal to find such code, as opposed to a script which does a bunch of other things as well.

Ghoulboy78: So in short, I would suggest extracting the item_consumed() func to some library, and importing form there, or signaling ot there, whatever.

programs/libraries/debug.sc
```diff
@@ -0,0 +1,127 @@
...
+        actuallyFormatted += format(_);
+      ,
+        actuallyFormatted += _;
+      );
+    );
+
+    allFormatted += actuallyFormatted + ' ';
+  );
+
+  print(entity_selector('@a'), allFormatted);
```
Ghoulboy78, line 116: I would suggest allowing to print to one specific player, but doing the following:
```
debug(...v) -> (
  if((type(v:0)=='entity' && v:0~'type'=='player') || (type(v:0)=='list' && all(v:0, type(_) =='entity' && _~'type'=='player')),
     player = v:0,
     player = player('*')
  );
  allFormatted = '';
  for (v,
    formatted = formatType(_);
    actuallyFormatted = '';
    for (formatted,
      if (type(_) != 'text',
        actuallyFormatted += format(_);
      ,
        actuallyFormatted += _;
      );
    );

    allFormatted += actuallyFormatted + ' ';
  );

  print(player, allFormatted);
);
```

This way if a player is specified as first arg, it will print just to that player (or player list), if not to all players as usual.

Xendergo: Ok, I've added that

programs/survival/compass.sc
Xendergo: So are you suggesting adding a new .sc file to listen for an item being consumed and letting other scripts import that?

Ghoulboy78: imo just forget about the whole importing business, just add the function and be done with it.

Xendergo: I'm not sure what you want me to do, the reason I'm signalling this as an event is so that other scripts that consume items can provide support for the persist option by signalling the event so the compass script can take care of replenishing the items if necessary. It's not really meant for other scripts to listen to.

Like for example, if someone was using the angel blocks script in a manhunt where the players spawn with cobblestone where the cobblestone never runs out, that wouldn't work when the player tries to place a block in the air, since the script removes stuff from the player's inventory, so to make it work, you could signal the event before removing the items, and the compass script will take care of checking if the persist option is used and replenishing the items if it is. I don't really think that importing a function would be the best way to do this.

If you're suggesting that I add a script that listens for when an item is used and signals an event, I'd be happy to do that, but it doesn't really seem that relevant to the compass script

Also if you're wondering why I didn't signal the event in the shoot_fireball script, it's because I don't have to, the compass script signals the event every time you use a fire charge, whether it's successful or not, I figured that just signalling the event anyways would be more efficient than doing ray tracing

Ghoulboy78: Yes, but you can't guarantee that someone has downloaded bot hthe compass AND shoot_fireball script.

Xendergo: I'm sorry, but I'm extremely confused. It doesn't matter if people would have both scripts loaded, you can signal an event even if there's no script to listen for it. There's also no function that removes items, the function is meant to give people the option to replenish items whenever they're consumed so they never run out

programs/survival/bee_healing.sc
```diff
@@ -0,0 +1,15 @@
+// Bees have an annoying habit of drowning themselves, so yeah
+__config()->{'scope' -> 'global'};
+
+begin()->(
+    bees = entity_selector('@e[type=bee]');
```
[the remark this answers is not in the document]
Xendergo: Good idea, I never knew that existed

programs/survival/bee_healing.sc
```diff
@@ -1,15 +1,24 @@
...
+
 begin()->(
-    bees = entity_selector('@e[type=bee]');
-    for (bees, 
-        health = _~'health';
-        if (health < 10,
-            modify(_, 'health', health+1);
+    for (global_bees,
+        if (_ ~ 'removed',
+            delete(global_bees, _i);
```
altrisi, line 9: Instead you could also (in the code below) handle the `on_removed` event to remove them.

programs/survival/compass.sc
Ghoulboy78: ye sorry I hadnt read code properly I meant that you should extract to a library

programs/survival/compass.sc
```diff
@@ -0,0 +1,613 @@
...
+            inventory_set(player, slot, 0);
+        );
+        slot = inventory_find(player, 'compass', slot+1);
+    );
+
+    bySlotNum = map(slots, _: 3); // All the inventory slots that are taken
+
+    // Find all the slots where the player has items that they respawn with
+    for (global_toKeep,
+        if (_: 3 == 'âny' || _: 3 == lower(player ~ 'name'), // It's impossible to use the character ̂  in a player name
```
Ghoulboy78, line 117: Can't u use null instead?

Xendergo: Yeah I could, I just didn't think of it at the time

programs/survival/bee_healing.sc
Xendergo: Ok, I've done that

programs/survival/compass.sc
Xendergo: Actually, I was testing it, and I got the error 'cannot reliably encode to a tag the value of 'null'' when saving the spawnWith data, so types are pretty important in this case

programs/survival/compass.sc
Xendergo: Ok, I've added a new script that makes signalling the event a proper function

programs/survival/compass.sc
Ghoulboy78: oh yeah, forgot...

Ghoulboy78: wait, try now with the new version? cos if not thats probably a bug

programs/survival/compass.sc
Ghoulboy78: do you want the app to have global or player scope?

Ghoulboy78 (commented):
It looks pretty good now, but in future it's best to stick to one app per pr, so it's less confusing to what's happening when

programs/survival/compass.sc
```diff
@@ -0,0 +1,613 @@
...
+        'time' -> {'type' -> 'float', 'min' -> 0.05, 'suggest' -> [0.5, 1, 10]},
+        'enabled' -> {'type' -> 'bool'},
+        'effect' -> {'type' -> 'effect'},
+        'duration' -> {'type' -> 'int', 'min' -> 0, 'max' -> 1000000, 'suggest' -> [15, 60, 1000000]},
+        'amplifier' -> {'type' -> 'int', 'min' -> 0, 'max' -> 255, 'suggest' -> [0, 1, 2, 3, 255]},
+        'dist' -> {'type' -> 'int', 'min' -> 16, 'max' -> 24000000, 'suggest' -> [8, 128, 7000, 24000000]},
+    }
+};
+
+toKeep = parse_nbt(read_file('toKeep', 'nbt'));
```
BisUmTo, line 47: It might be useful to use nbt_storage() instead 🤔

programs/survival/compass.sc
```diff
@@ -0,0 +1,613 @@
...
+    );
+);
+
+global_playerPos = {};
+
+changeCompassPos() -> {
+    // Store the current locations of all the players
+    updatePlayerPosition();
+
+    players = entity_selector('@a');
```
BisUmTo, line 178: Why do you sometime use entity_selector('@a') and sometimes player('*')?

BisUmTo (approved):

programs/survival/compass.sc
Xendergo: The only time I could see `player('*')` being used was the part of debug.sc that ghoulboy wrote and I modified a bit, [linked here](https://github.com/gnembon/scarpet/pull/177#discussion_r575848600). I use `entity_selector('@a')` because it gets players from every dimension, which is usually what I want

programs/survival/compass.sc
Xendergo: Looking through the documentation, I feel like it'd probably be best for me to refactor it to use global scope, I didn't really understand the difference at the time

programs/survival/compass.sc
Xendergo: That's probably a good idea

programs/libraries/debug.sc
```diff
@@ -0,0 +1,129 @@
+// Adds a debug function which is equivalent to the print function, but prints stuff different colors depending on type. Avoids confusion between null and 'null' and between maps, strings, and nbts
```
gnembon, line 1: 'debug' is a very generic function. While I understand the overal utility of this method, would suggest naming it more specific with something that is more recognizable: pretty_print

programs/libraries/README.md
```diff
@@ -0,0 +1,40 @@
+# Libraries
+Programs for adding helpful features not in native scarpet
+
+## Libraries in alphabetical order with creator:
+
+### compass_persist.sc:
+#### By Xendergo
```
gnembon, line 7: feels very specific to that compass app - not sure why this would need to be a generic use library.

programs/libraries/debug.sc
Xendergo: Ok, I did this

programs/libraries/README.md
Xendergo: Idk either, ghoulboy asked for this, although I could've easily misunderstood: https://github.com/gnembon/scarpet/pull/177#discussion_r576392390

I'd be happy to remove it

Ghoulboy78: What were saying is forget about the library functions and event calls across apps. ALl the apps should be able to run in their own environment for ease of use, so just drop all of that.

Xendergo: Ok, I've removed it then

Ghoulboy78: Ok, can you remove it from the ReadMe then?

Ghoulboy78: oh wait you did.

programs/libraries/pretty_print.sc
```diff
@@ -0,0 +1,129 @@
...
+// map - red
+// iterator - Darker yellow
+// function - magenta
+// task - grey
+// block - green
+// entity - Darker red
+// nbt - cyan
+// text - Unchanged, usually formatted already
+
+formatType(v) -> (
```
Ghoulboy78, line 17: generally scarpet names are snake case, not camel case (so `format_type` instead of `formatType`). I believe this is a convention across statically typed languages, though I may be mistaken. Certainly in scarpet, in-built functions must be snake-case

programs/survival/bee_healing.sc
```diff
@@ -0,0 +1,40 @@
+// Bees have an annoying habit of drowning themselves, so yeah
+__config()->{'scope' -> 'global'};
+
+global_incrementer = 0;
+global_bees = {};
+
```
Ghoulboy78, line 6: Why need a map and incrementer? If you're going to iterate over them all anyways, can't you just use a list? Also couldn't you use an entity event which schedules 800 ticks later?

programs/libraries/pretty_print.sc
Xendergo: Yeah you're right, I should change it.

programs/survival/bee_healing.sc
Xendergo: If bees don't die/unload in the reverse order they were loaded, then you would have to delete items from the middle of  the list, which would change the indexes of all the other bees in the list, which would break everything. That's why I have to use a map

I suppose you could call a recursively scheduling function for every bee that spawns, but I don't see how that would be a better implementation

Ghoulboy78: Well the recursively calling function would run for each bee, allowing for cleaner code, as you don't need to keep track of each bee, just use the entity_load_handler to assign a function like `track_health(bee)` to each bee, which incrementes the health of the bee it it's less than 10, and then at the end does `if(!bee~'removed', schedule(800, 'track_health', bee))` to call itself again if the bee is alive. In fact, that call should be put at the beginning of the function, so the whole thing looks like this:
`bee_healing.sc`:
```
__config()->{'scope'->'global'};

track_health(bee)-> if(!bee~'removed',
    health = bee~'health';
    if(health < 10, modify(bee, 'health', health+1));
    schedule(800, 'track_health', bee)
);

entity_load_handler('bee', 'track_health');
```
That would work without any bee tracking shenanigans and no iterations etc. It just assigns the event, and then leaves the bees be. (pun intended, I have now descended to the level of Cory Scheviak in terms of horrible bee puns...)

programs/libraries/pretty_print.sc
```diff
@@ -0,0 +1,129 @@
...
+);
+
+concat(a, b) -> (
+  ret = copy(a);
+
+  for (b,
+    put(ret, null, _);
+  );
+
+  return(ret);
```
Ghoulboy78, line 128: no need for return call here, justput `ret`

programs/libraries/pretty_print.sc
```diff
@@ -0,0 +1,129 @@
...
+    c: 0 = 'g '+v;
+  , t == 'block',
+    c: 0 = '#00ff00 '+v;
+  , t == 'entity',
+    c: 0 = '#ff2222 '+v;
+  , t == 'text',
+    c: 0 = v;
+  );
+
+  return(c);
```
Ghoulboy78, line 93: No need for return call here either

programs/survival/compass.sc
Ghoulboy78: My bad, the correct usage would be `player('all')` to get all players from all dimensions in the fastest way possible. `entity_selector('@a')` obviously works, but is slower, as it has to go through the entire rigmarole of the vanilla entity querying thing ma bob, then returns the players, whereas scarpet just reads the contents of an internal Java variable.

programs/libraries/pretty_print.sc
```diff
@@ -0,0 +1,129 @@
...
+// map - red
+// iterator - Darker yellow
+// function - magenta
+// task - grey
+// block - green
+// entity - Darker red
+// nbt - cyan
+// text - Unchanged, usually formatted already
+
+format_type(v) -> (
```
Ghoulboy78, line 17: Couldn't you put this in a map? like:
```
type_format_colour_map = {
   'null' -> 'p ',
   'number' -> '#0080ff '
}
```
etc?

Ghoulboy78: Ah wait nvm I understand why that wouldnt work for lists and maps, yeah. I mean, you could technically do something with lambda functions etc, but that would be more bulky and less useful.

programs/survival/bee_healing.sc
Ghoulboy78: @Xendergo I would recommend replacing bee_healing.sc with what I've posted above, as it does the exact same job without iterating over all the bees each time, instead letting each bee keep track of itself.

programs/survival/compass.sc
```diff
@@ -0,0 +1,604 @@
+// If you want your script to work with the persist flag, signal the relevant scarpet event
+// For example: signal_event('player_places_block', player, player, l('acacia_door', 16, null), 'mainhand', block(0, 0, 0));
+
+__config()->{
+    'scope' -> 'global',
+    'commands' -> {
+        'giveCompass' -> _() -> giveCompass(player()),
```
Ghoulboy78, line 7: Again here, try use snake case for functions etc.

programs/survival/compass.sc
```diff
@@ -0,0 +1,604 @@
...
+        'track' -> _() -> print(player(), format('l Make the compass you\'re holding track a player')),
+        'spawnWith' -> _() -> print(player(), format('l Spawn with a particular item or effect when anyone or a specified player dies')),
+        'dontSpawnWith' -> _() -> print(player(), format('l Stop spawning with a particular item or effect')),
+        'logOnTrack' -> _() -> print(player(), format('l Send a message when someone tracks you: ', 'lb '+if(global_logOnTrack, 'true', 'false'))),
+        'maxCompasses' -> _() -> print(player(), format('l The maximum amount of compasses one can give themselves: ', 'lb '+global_maxCompasses)),
+        'noPortalBreaking' -> _() -> print(player(), format('l Whether to allow breaking portals: ', 'lb '+if(global_noPortalBreaking, 'true', 'false'))),
+        'maxNetherTravel' -> _() -> print(player(), format('l The maximum distance players can travel in the overworld thorugh the nether: ', 'lb '+(global_maxNetherTravel*8))),
+    },
+    'arguments' -> {
+        'player' -> {'type' -> 'string', 'suggester' -> _(args) -> entity_selector('@a'), 'case_sensitive' -> false},
```
Ghoulboy78, line 36: You can replace `entity_selector('@a')` with `player('all')`. It is true that that may conflict with a player whose name is 'all', but 3-letter names and less are no longer a thing, and player('all') is hardcoded to work and capture all players in all dimensions.

Xendergo: I did some testing on my machine, and found that `entity_selector('@a')` was consistently much faster than `player('all')`, however, they're both so fast that I'm not sure it really matters anyways

```
/script run for(range(1000000), _)                                    // around 100ms
/script run for(range(1000000), entity_selector('@a'))                // around 300ms
/script run for(range(1000000), player('all'))                        // around 450ms
```

I'd be curious to see what your results are, or if you think this is even a fair test

programs/libraries/pretty_print.sc
Xendergo: Ok, I've removed that

programs/libraries/pretty_print.sc
Xendergo: removed that too

programs/survival/bee_healing.sc
Xendergo: Ok, I've replaced it, that definitely looks like a nicer solution

programs/survival/compass.sc
Xendergo: Ok, I've switched what I could find to snake case. From what I can tell it still seems to work

Imo, using snake case for the commands seems kinda pointless

programs/survival/compass.sc
Ghoulboy78: better to do:
```
/script run profile_expr(_)
/script run profile_expr(entity_selector('@a'))
/script run profile_expr(player('all'))
```
cos that (i think) runs it better to test it to your system.
I'd check those, and then see.

Xendergo: The `entity_selector('@a')` one was still faster for me

Firigion (approved):
Looks good to me!

altrisi:
Sorry for the long delay btw, we've been quite slow in this for sure. Hopefully next time the same won't happen! Also feel free to not apply all the comments given, some are just so you can know for next time, but not requirements!

PR #178, opened 2021-01-20
witch_hut became swamp_hut sometime during 1.16.x

random-j-farmer:
By leaving the old line, it should still work in earlier versions.


gnembon:
why don't you add that around line #808 with other backwards compatibility things.

random-j-farmer:
I did check that it still finds witch huts, but only in 1.16.4

BisUmTo (approved):

replaceitem (approved):

Issue #180, opened 2021-02-09
cam.sc does not teleport back to correct dimension on a server


TheCatSaber:
How to reproduce:

Install cam.sc on a server. (Issue did not occur on singleplayer).

Go into survival.

Run `/cam`.

Run `/execute in minecraft:the_nether tp @p 0 128 0` to teleport to the nether.

Run `/cam` again.

You should now come back out in the nether.

TheCatSaber:
My interim measure to to replace lines 36-38 with this, to make it the player's issue.
```
         // Only make survival if in same dimension
         if (p~'dimension'==config:'dimension',
            __remove_camera_effects(p);
            __restore_player_params(p, config);
            __remove_player_config(p~'name'),
            // Otherwise say wrong dimension
            p~'dimension'!='dimension',
            print('You must be in the same dimension you took off in: '+config:'dimension'+'.')
         )
```

Ghoulboy78:
hmm, true. Thought it had been fixed...

altrisi:
Oh, only on a server. That's why I couldn't reproduce it.

altrisi:
Nope, still can't reproduce, works for me. Are you for any chance running Carpet TIS Addition with `opPlayerNoCheat` rule?

The cam app currently relies on using the `tp` vanilla command to change dimensions (since it's not yet supported natively in Scarpet, I'm in the process of adding it), and that rule blocks Scarpet from running the command.

TheCatSaber:
I am not running the opPlayerNoCheat, and I am unsure why it won't work, given the command console can still do ```/execute in minecraft:dimension run tp player ~ ~ ~```, and it works correctly. I intend to see if changing the command to tping to 0 128 0 works when I get the chance. I also noticed at least 1 other person had this issue on scicraft discord recently.

altrisi:
Hm... Maybe try changing the `commandScriptACE` rule to a higher value, though it should already be 2 ( == `ops`, min permission to run `tp`) by default. That's the only way I've been able to not tp out of the nether.

TheCatSaber:
Solved: The carpet.conf started with #locked, overriding the commandScriptACE, therefore you have to set commandScriptACE equal to 2 in carpet.conf. Will make a PR to put in comments of cam.sc, the readme, and also edit the fabric-carpet wiki.

altrisi:
I'll keep this open while one of the fixes is done, so others can see the workaround.

PR #183, opened 2021-02-17
Add info to fix dimension teleporting issues.

TheCatSaber:
See #180 for more details. Will update fabric-carpet wiki to reflect this info and line number changes after merged.

### DIFF
--- a/programs/survival/README.md
+++ b/programs/survival/README.md
@@ -33,6 +33,8 @@ Various scripts that modify various game elements, often replicating popular mod
 #### By Gnembon
     As /c /s used to be in the olden 1.15- days, just configurable, safe with anti-cheat mechanisms, saving player landing location in 
     app nbt's for future use. The whole shabang.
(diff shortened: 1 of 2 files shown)


altrisi:
Oh, actually that may be a bug in fabric-carpet.

This happens because Carpet automatically disables all commands by default in a `locked` config file, and the way a "command" is defined is by checking whether rules have the `COMMAND` category, which is a categiry also applies to this.

I don't think that is intended...

TheCatSaber:
Maybe the default `commandScriptACE` should be 2 or a new value for server only (if that's possible), even when locked, because people won't be able to run scarpet commands with `commandScript` set to an appropriate value.

If you make an issue on the fabric-carpet repo, I'll close this PR, and you should also let the scarpet helpers on discord know that to fix the issue of dimension teleporting in cam.sc, you must put `commandScriptACE 2` in carpet.conf.

PR #187, opened 2021-02-24
Delete camera.sc because already in Fabric Carpet

DragonEggBedrockBreaking:
Delete camera.sc because already in Fabric Carpet, and the one in fabric carpet seems to have been updated much more recently and the one in this repository seems a bit abandoned.


BisUmTo:
I think it's better to update the script in this repository too or add a link somewhere to the built-in sources.

DragonEggBedrockBreaking:
> I think it's better to update the script in this repository too or add a link somewhere to the built-in sources.

I made #188 for this then. I will keep this open so Gnembon can choose which to merge.

PR #189, opened 2021-02-24
Upgrade deprecated usages in storage drawers

### LINKED ISSUE #186: storage_drawers uses deprecated APIs
storage_drawers.sc uses __on_tick_nether() and __on_tick_ender(), which are deprecated.  Given that this is one of gnembon's  own scripts, and is likely to be used as reference, it should be updated.

altrisi:
Fixes #186.

- Removes usages of `__on_tick_nether` and `__on_tick_end` and changes them to the current `__on_tick(in_dimension('...',...))`
  - As a bonus, this should also now work for non-vanilla dimensions by just specifying them in the list
- Removes useless things from the `__config`
- Imports `distance_sq` from Carpet's `math` library instead of recreating it
- It seems it also adds the newline at the end. Didn't know it wasn't there, it's nice that it got added automatically
- Oh, and it also removes calling `__on_start()` since that is now done by Carpet (same way since it's global), else would be called twice

I didn't touch the command system since it's basically the root command. Also didn't upgrade anything else because just in case, since I'm not familiar with the app and I don't want to break it without the need to or create any unnecessary diff. These were the basics.

Note: I still haven't tested this. I'll probably do it Soon™.

### DIFF
--- a/programs/survival/storage_drawers.sc
+++ b/programs/survival/storage_drawers.sc
@@ -6,7 +6,7 @@ global_dims = l('overworld', 'the_nether', 'the_end');
 
 __config() ->  {
 	'scope' -> 'global'
-}
+};
 
 //displays all controller positions and detailed content of the closest storage
 __command() ->
--- a/programs/survival/storage_drawers.sc
+++ b/programs/survival/storage_drawers.sc
@@ -23,7 +23,7 @@ __command() ->
 	in_dimension( current_player,
 		loaded_controllers = sort_key(
 			filter(global_controllers:current_dimension(), loaded(_)), 
-			__euclidean_sq(_,player_pos)
+			_euclidean_sq(_,player_pos)
 		);
 		if (loaded_controllers,
 			closest_controller = loaded_controllers:0;
--- a/programs/survival/storage_drawers.sc
+++ b/programs/survival/storage_drawers.sc
@@ -1,19 +1,12 @@
+import('math', '_euclidean_sq');
+
 global_controllers = null;
 global_banned_blocks = m('hopper', 'dropper', 'dispenser');
 global_dims = l('overworld', 'the_nether', 'the_end');
 
-// must turn
-// /carpet setDefault scriptsAutoload true
-// for the scripts to start with the world automatically
-__config() -> 
-(
-	m(
-		l('stay_loaded', true),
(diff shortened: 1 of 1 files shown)


BisUmTo (approved):

replaceitem (approved):

programs/survival/storage_drawers.sc
```diff
@@ -1,19 +1,12 @@
...
-	m(
-		l('stay_loaded', true),
-		l('scope', 'global')
-	)
-);
-
-__distance_sq(v1, v2) -> reduce(v1-v2,_*_+_a,0);
+__config() ->  {
+	'scope' -> 'global'
+}
```
gnembon, line 9: semicolon ;)
```suggestion
};
```

gnembon:
Caught a missing semicolon. Probably needs a /script load quick test

programs/survival/storage_drawers.sc
```diff
@@ -132,8 +125,6 @@ __load_controllers(tag) ->
 	)
 );
 
-__on_start();
-
```
Ghoulboy78, line 136: Wait, if you remove this call, then that function is never called mate

Ghoulboy78: And that code never runs afaik

altrisi: That is now an event in Carpet (as specified in the pr body).

Ghoulboy78: oh Im dumb I didnt read properly lol

Ghoulboy78:
BTW, would you be against adding more specific debug commands or smth to make it easier to navigate this? I feel it may make it easier to see what's happening, especially for new people who may not have seen the video or understood it fully. Nothing major I don't think, but just a thought?

altrisi:
I'd first merge this to get it working without the annoying red bold **"uses deprecated functions"** and then add more things, though it has to be tested first (I'm recently being lazy and not even loading this script, lol).

BTW since here PRs will usually be quite mixed and don't really make sense to keep the entire history (maybe there are a lot of ammendments to an app or whatever) I think here it would be a good idea to squash PRs, to keep a cleaner, just PRs, history (IMO).

altrisi:
Tested, now works. Merging

PR #191, opened 2021-02-26
Added pickup_hammer.sc

rv3r:
Requested by Gaduji on the SciCraft discord, figured more people might want this


rv3r:
Kollum informed me that a 1 tick delay in auto_pickup does the trick, no need to combine

PR #196, opened 2021-03-21
Add flower forest flowers script

manyrandomthings:


altrisi (commented):
Overall looks good, will approve after testing if another look at it doesn't find other simple improvements.

However I'd suggest you remove the `return` statements inside of functions called by loops, since those are quite slow in Scarpet. You don't need them at the end of functions since functions already return whatever is in the last statement by default (the ones that return inside of an `if` statement would need an `else` block as an equivalent to the return in the exterior). Maybe keep the ones that would get less readable that way, since it isn't really a vital/always looping script, but should improve performance when filling large amount of blocks.

BisUmTo (approved):

BisUmTo:
I love this script!
I spent so much time with dispensers /fill command

programs/utilities/flower.sc
```diff
@@ -0,0 +1,136 @@
...
+  x = position : 0;
+  z = position : 2;
+
+  // gets simplex noise for given  coordinate with seed 2345 (from vanilla code)
+  val = simplex(x / 24, z / 24, null, 2345);
+  // makes sure output is between 0 and 0.9999
+  val_clamped = min(max(val, 0), 0.9999);
+  flower_id = floor(val_clamped * length(global_FLOWERS));
+
+  return(flower_id);
```
BisUmTo, line 81: it is not essential, it will return the value of the last line anyway

programs/utilities/flower.sc
```diff
@@ -0,0 +1,136 @@
...
+  'stay_loaded' -> true,
+  'commands' -> {
+    '<pos> print' -> _(position) -> print(get_flower_at(position)),
+    '<pos> create <radius>' -> 'flower_area',
+    '<pos> create <radius> only <flower>' -> 'filtered_flower_area'
+  },
+  'arguments' -> {
+    'pos' -> {
+      'type' -> 'pos'
+    },
```
Ghoulboy78, line 25: Not necessary, scarpet figures it out on its own

programs/utilities/flower.sc
```diff
@@ -0,0 +1,136 @@
...
+);
+
+get_flower_at(position) -> (
+  flower_id = _get_flower_id_at(position);
+  return(global_FLOWERS : flower_id);
+);
+
+// check for flower forest biome
+_is_flower_forest(position) -> (
+  return(biome(position) == 'flower_forest');
```
Ghoulboy78, line 91: try to avoid unnecessary return calls, cos they tend to be slow in scarpet specifically. In this case, you can just do
```
_is_flower_forest(position) ->
  biome(position) == 'flower_forest';
```

programs/utilities/flower.sc
```diff
@@ -0,0 +1,136 @@
...
+  // makes sure output is between 0 and 0.9999
+  val_clamped = min(max(val, 0), 0.9999);
+  flower_id = floor(val_clamped * length(global_FLOWERS));
+
+  return(flower_id);
+);
+
+get_flower_at(position) -> (
+  flower_id = _get_flower_id_at(position);
+  return(global_FLOWERS : flower_id);
```
BisUmTo, line 86: you can remove the return word, if you want.

BisUmTo (approved):

Ghoulboy78 (approved):

PR #197, opened 2021-03-29
update to villager POI system

sanketc001:


Firigion (approved):
Thanks

Ghoulboy78 (approved):

altrisi:
Can't really find any reference to emeralds in the script, though.

Ghoulboy78:
OH gosh, yes, you're right mate. Whoops.

altrisi:
Only reference I can find to emeralds seems to be from the `track` command in base Carpet, which seems to do exactly that. At

https://github.com/gnembon/fabric-carpet/blob/3bf3c3116071fa4917f034e01c9ffa999b3edd2a/src/main/java/carpet/mixins/VillagerEntity_aiMixin.java#L122

altrisi:
Hello @sanketc001, thanks for contributing!

I think you confused the `villager_poi.sc` app from this repo with the features of the `/track` command in Carpet, is that possible? If you wanted to contribute documenting that command, you should instead go to the [Commands](https://github.com/gnembon/fabric-carpet/wiki/Commands) page of the Carpet wiki, where the track command is also undocumented.

Thanks!

_If I'm incorrect in something, please reply and I'll reopen this issue._

PR #205, opened 2021-04-23
Added bucketstack.sc survival script and description

rv3r:
Wrote this script a few months ago from a request but forgot to make a pr to add it to the repo


programs/survival/bucketstack.sc
```diff
@@ -0,0 +1,129 @@
...
+(
+	//figure out how many items to move
+	//and move them
+	move = min(max_stack - slot_count,count);
+	inventory_set(player,slot,slot_count + move,item);
+	count = count - move;
+	//if all possible slots are full, leave whatever is left on the ground
+	//modified line 90 of gnembon's shulkerboxes.sc to change the number of items in the item entity
+	modify(entity,'nbt_merge','{Item:{Count:' + count + 'b}}');
+	//line 68 of gnembon's shulkerboxes.sc to play the item pickup sound
```
Ghoulboy78, line 65: Would this play the animation properly though? I think gnembon added that to shulkerboxes.sc as well...

programs/survival/bucketstack.sc
```diff
@@ -0,0 +1,129 @@
...
+//WRITTEN BY FIRIGION
+
+__on_player_right_clicks_block(player,item_tuple,hand,block,face,hitvec) ->
+(
+  if(hand == 'mainhand',
+		slot = player~'selected_slot',
+		//-1 is offhand
+		slot = -1;
+	);
+  if(item_tuple:0~'_bucket' && !item_tuple:0~'milk',
```
Ghoulboy78: Why need to check that it's not milk?

programs/survival/bucketstack.sc
```diff
@@ -0,0 +1,129 @@
...
+  if(item_tuple:0~'_bucket' && item_tuple:0~'milk',
+		schedule(0,'test_bucket_used',player,hand,item_tuple,slot);
+	);
+);
+
+test_bucket_used(player, hand, item_tuple, slot) ->
+(
+	l(item,count,nbt) = item_tuple;
+	//trigger only if bucket was used
+  if(query(player,'holds',hand):0 == 'bucket' || item~'milk',
```
Ghoulboy78: But don't we already know by this point that a bucket was used?

programs/survival/bucketstack.sc
```diff
@@ -0,0 +1,129 @@
...
+
+test_bucket_used(player, hand, item_tuple, slot) ->
+(
+	l(item,count,nbt) = item_tuple;
+	//trigger only if bucket was used
+ 	if(inventory_get(player,slot):0 == 'bucket' || item~'milk',
+		if(count == 1,
+			//do nothing if it was the last bucket in the stack
+			return();
+		);
```
Ghoulboy78, line 107: Maybe you could add the count!=1 to the if statement to avoid return() call

programs/survival/bucketstack.sc
```diff
@@ -0,0 +1,129 @@
...
+  	if(item_tuple:0~'_bucket' && item_tuple:0~'milk',
+		schedule(0,'test_bucket_used',player,hand,item_tuple,slot);
+	);
+);
+
+test_bucket_used(player, hand, item_tuple, slot) ->
+(
+	l(item,count,nbt) = item_tuple;
+	//trigger only if bucket was used
+ 	if(inventory_get(player,slot):0 == 'bucket' || item~'milk',
```
Ghoulboy78, line 103: Don't we already know that the player has bucket by now?

programs/survival/bucketstack.sc
```diff
@@ -0,0 +1,129 @@
...
+
+//only for milk
+__on_player_finishes_using_item(player,item_tuple,hand) ->
+(
+	if(hand == 'mainhand',
+		slot = player~'selected_slot',
+		//-1 is offhand
+		slot = -1;
+	);
+  	if(item_tuple:0~'_bucket' && item_tuple:0~'milk',
```
Ghoulboy78, line 94: y need to check for not milk?

Ghoulboy78 (commented):
Pretty good, just got a few recommendations.

programs/survival/bucketstack.sc
rv3r: This is needed on line 81 as right-clicking on a block can be done with milk actually without using the milk, creating an unnecessary call to test_bucket_used() if we don't catch it early. I'll change line 94 to just check for 'milk_bucket'.

programs/survival/bucketstack.sc
rv3r: Will do.

programs/survival/bucketstack.sc
rv3r: Good catch. Just implemented it in a similar style.

programs/survival/bucketstack.sc
rv3r: We previously checked that the item that was being used was a bucket variant. At this point, the bucket has been used and the slot _should_ still have an empty bucket or nonzero milk buckets. In the event that a player is swapping hands quickly or using Tweakeroo to rapidly change the item in a slot, we recheck before setting a bucket to prevent bucket duping.

rv3r:
Is the new-app label something I should've added or is that something that only the reviewers can do?

Ghoulboy78 (approved):
Well, ig this is good enough for me, it works I suppose. I'll give it a more thorough check later to try and see if I can break it.

programs/survival/bucketstack.sc
```diff
@@ -0,0 +1,133 @@
...
+		l('water',1),
+		l('lava',1),
+		l('milk',1),
+		l('pufferfish',1),
+		l('salmon',1),
+		l('cod',1),
+		l('tropical_fish',1),
+		l('axolotl',1)
+	);
+	return(
```
Firigion, line 15: No need for a `return()` call here, pretty sure. And you can use the more readable `[·]` and `{·}` calls intead of `l(·)` and `m(·)`.

programs/survival/bucketstack.sc
```diff
@@ -0,0 +1,133 @@
...
+	if(entity~'pickup_delay' == 0,
+		l(item,count,nbt) = entity~'item';
+		if(item~'_bucket',
+			//if it's an item we're picking up that is a non-empty bucket
+			//first determine the max stack size
+			type = null;
+			while(type == null,length(global_buckets),
+				type = item~(keys(global_buckets):_);
+			);
+			max_stack = global_buckets:type;
```
Firigion, line 34: why not `max_stack = global_buckets:item;`? If you want to avoid conflicts with modded bucket items, you can set a default stack size like so:
`max_stack = global_buckets:item || 1;`

Maybe I'm just misunderstanding wht this does.

programs/survival/bucketstack.sc
```diff
@@ -0,0 +1,133 @@
...
+			//then find a non-full slot to put it in
+			//modified lines 103 and 104 of gnembon's shulkerboxes.sc to find a non-full slot
+			slot = -1;
+			while(count && (slot = inventory_find(player,item,slot+1)) != null && slot < 36,inventory_size(player)-5,
```
Firigion, line 38: Itn's this avoiding the offhand too?

programs/survival/bucketstack.sc
```diff
@@ -0,0 +1,133 @@
...
+				if(slot_count < stack_limit('bucket') && bucket_slot != slot,
+					inventory_set(player,bucket_slot,slot_count + 1,'bucket');
+					return();
+				);
+			);
+			//hm better search for an empty slot if the above code didnt find anything
+			while((bucket_slot = inventory_find(player,null)) != null && bucket_slot < 36,inventory_size(player)-5,
+				inventory_set(player,bucket_slot,1,'bucket');
+				return();
+			);
```
Firigion, line 128: You could bastract this, since you also use it a few lines up and make both codes a bit more readable, no?

Firigion (commented):
I found two bugs, beyond te minor comments in this review:
- Doesn't work in creative mode (I don't think that's an issue, but it should be stated somehwere)
- There are missing calls to `__on_player_uses_item()` and `__on_player_interacts_wth_entity()` that should also grab the potentially generated bucket (don't forget goats for 1.17+ milk) and stacks it. Right now, it just drops to the ground.

rv3r:
You're right that it won't work in creative mode; I'll note that.

programs/survival/bucketstack.sc
rv3r: As far as I'm aware, no item ever goes into the offhand on pickup, not even a shield. Thus, the offhand should be avoided.

Ghoulboy78:
(the commit I made was just resolving a conflict)

programs/survival/bucketstack.sc
rv3r: Updated it to loop across the same code, first looking for stacks of buckets, then looking for empty slots.

programs/survival/bucketstack.sc
```diff
@@ -0,0 +1,131 @@
...
+					slot_count = inventory_get(player,bucket_slot):1;
+					//dont want to overfill a slot or overwrite the original slot
+					if(slot_count < stack_limit('bucket') && bucket_slot != slot,
+						inventory_set(player,bucket_slot,slot_count + 1,'bucket');
+						return();
+					);
+				);
+			);
+			//if there were absolutely no places to put the bucket, just spawn a bucket item
+			spawn('item',pos(player)+l(0,((player~'eye_height')/2),0),nbt(m(l('Item',m(l('id','"minecraft:bucket"'),l('Count','1b'))),l('PickupDelay',0))));
```
Ghoulboy78, line 128: You can also replace the nbt() call with a string, imo that's easier to read than this with all the m() and l() calls

programs/survival/bucketstack.sc
rv3r: It should work, but now I get an `unknown_item exception` because apparently the string `'bucket'` from line 113 is passed to `inventory_find` as 0. No clue what the problem is.

programs/survival/bucketstack.sc
rv3r: I agree; it looks better now.

programs/survival/bucketstack.sc
rv3r: Fixed. Thanks

programs/survival/bucketstack.sc
Firigion: Not even to stack?

programs/survival/bucketstack.sc
Firigion: `while` might be using the `_` it itslef defines, overwriding the outer `_` defined by `for`. Try doing `current_item = _` before entering the `while` loop.

rv3r: Yep that was it.

programs/survival/bucketstack.sc
rv3r: You were right; if there was already a partial stack, the offhand will take the items that are picked up. Now the script matches the vanilla behavior.

programs/survival/bucketstack.sc
rv3r: Return call has been removed

rv3r: While I don't mind changing these, keep in mind that readability of  `[]` and `{}` over `l()` and `m()` is not purely objective. If we have a standard for this repo, I'm happy to follow it.

rv3r:
Stacking of buckets when the player generates a new bucket(filling with a liquid or block, bucketing an entity) is currently being worked on. Please do not approve of this pull until then.

rv3r:
Any thoughts on whether unique tropical fish and axolotl buckets should stack? I say no.

rv3r:
After a large number of improvements and changes today, I think this is in a state to be reviewed again before addition to the repo.

programs/survival/bucketstack.sc
```diff
@@ -0,0 +1,179 @@
...
+		l('water',1),
+		l('lava',1),
+		l('milk',1),
+		l('pufferfish',1),
+		l('salmon',1),
+		l('cod',1),
+		l('tropical_fish',1),
+		l('axolotl',1),
+		l('powder_snow',1)
+	);
```
BisUmTo, line 15: I think that an ingame way to change this config would be really cool
`/bucketstack max <item> set <count>`
`/bucketstack max <item> get`

rv3r: Done

BisUmTo (approved):
it looks fine to me 😉

altrisi:
Does a player running the command update the sizes for the rest? Given it's player scoped it may not (now: should it?). I'd also make the command to change the stack sizes be op only. But I don't really want to delay this even more.

rv3r:
Since the stack sizes are stored in app data and not in a player scoped variable, the stack sizes do change for all players.

rv3r:
Is there a way to make custom commands op only?

altrisi:
> 
> 
> Is there a way to make custom commands op only?

Yes, `command_permission` in [`__config()`](https://github.com/gnembon/fabric-carpet/blob/master/docs/scarpet/Full.md#app-config-via-__config-function).

rv3r:
I guess it seems sensible to make the command op only. I might not use that method since it would also prevent any player from inquiring about the current stack size. Perhaps I can get the op status/permission level of the player and use that to prevent changes.

programs/survival/bucketstack.sc
```diff
@@ -0,0 +1,223 @@
...
+(
+	m(
+		l('scope','player'),
+		l('stay_loaded',true),
+
+		l('commands',
+			m(
+				l('<bucket>','__print'),
+				l('<bucket> <stack>','__change')
+			)
```
Firigion, line 11: how about adding an `/bucketstack all <size>` for a global stack size for all bucket types?

programs/survival/bucketstack.sc
```diff
@@ -0,0 +1,223 @@
...
+__initialize() ->
+(
+	if(!load_app_data(),
+		buckets = m();
+		for(filter(item_list(),_ ~ '_bucket') - '_bucket',
+			buckets:_ = 1;
+		);
+		store_app_data(buckets);
+	)
+);
```
Firigion, line 90: I'm not sure about removnig the in-app configuration of default stack sizes. It makes it easier to use, imo (just edit the values when savign the app, no need to run eight different commands in-game). I'd say, leave the in-app configured values and override them with the command defined ones, if given.

Firigion (commented):
I can't test the app rn, but this are my comments after reading over the thing. This app is shaping nicely :)

programs/survival/bucketstack.sc
rv3r: It makes it easier only for first time use and only if you want to change many of the stack sizes at once. I don't know what you think, but I suspect that most people would change water, lava, powder snow, and maybe milk or axolotls. I don't think that many people would want to change every single stack size, so only a few commands would be required. Matching tropical fish nbt is very hard and unless you need a lot of the other fish, there's no need to change the default stack size.

I can put the configuration map back in, but I would make sure that the app data values override it.

programs/survival/bucketstack.sc
rv3r: Seems fine to me.

programs/survival/bucketstack.sc
Firigion: Tropical fish buckets are used to breed axolotl, for instance. And I don't think it would be unresonable to think people would want all buckets to stack to the same ammount. That being said, It's enough for me with the `all` command added and, ultimately, it's your app. I'm not gonna block it because it's not the way I'd do it. When I can, I'll test it in game and approve if I don't find bugs :)

rv3r: I can understand wanting tropical fish buckets to stack, but I'm only allowing that when nbt matches because otherwise players will be able to dupe any fish type.

Firigion: Yes, yes, I agree with that.

Firigion:
Found a minor bug:
image
I'd like it for it to list all stack sizes, since it's a good way to check the current state of all stacks.

Firigion:
Another bug: using a bucket (both on an entity and a block) will frist look for an empty slot to put the bucket in, then stack the buckets, resulting in the inventory fillin gup before stacking.

Firigion:
Idk if it's important enough and/or if it's fixable, but picking up a fish with a bucket doesn't give you the advancement for doing it for the first time (tactical fishing) unless the bucket gets dropped because the inventory is full.

Also, on a more relevant not. tropical fish buckets that should stack (same NBT) stack only whe picking it up, not when right clicking a fish. Axolotl never stack, I guess they hace more complex nbt that doesn't let them stack.

Firigion:
Sometimes, when picking up lava, the new bucket is dropped instead of being inserted into the inventory. I'm reporting these bugs without doing code analysis, I can't rn.

rv3r:
> I'd like it for it to list all stack sizes, since it's a good way to check the current state of all stacks.

fixed in 8399ea1

rv3r:
> Idk if it's important enough and/or if it's fixable, but picking up a fish with a bucket doesn't give you the advancement for doing it for the first time (tactical fishing) unless the bucket gets dropped because the inventory is full.

Describe the exact situation for this please. If I pick up a fish, revoke the advancement, and pick up another fish, the bucket stacks and I get the advancement. I'm picking up the first fish so the item has the fish health in nbt. Getting it from the creative menu, `inventory_set()`, or `/give` does not give that nbt and they won't stack.

rv3r:
> Also, on a more relevant not. tropical fish buckets that should stack (same NBT) stack only whe picking it up, not when right clicking a fish. Axolotl never stack, I guess they hace more complex nbt that doesn't let them stack.

fixed in b3f8134

rv3r:
> Sometimes, when picking up lava (edit: or other things), the new bucket is dropped instead of being inserted into the inventory.
fixed in 27522ff

rv3r:
> Another bug: using a bucket (both on an entity and a block) will frist look for an empty slot to put the bucket in, then stack the buckets, resulting in the inventory filling up before stacking.

In these cases, I'm allowing vanilla behavior to take over. Yes, I can override this, but I wasn't concerned about it and I've been fixing everything else lately.

Firigion:
Awsome, I'll test again when I have some time!

Firigion:
Picking up a tropical fish now sometimes dupes the fsh bucket.

Firigion:
> Another bug: using a bucket (both on an entity and a block) will frist look for an empty slot to put the bucket in, then stack the buckets, resulting in the inventory filling up before stacking.
Not fully fixed: if the app finds an empty slot before a slot with the bucket type, then it puts the items in the bucket slot, instead of stacking it.

Firigion:
Misclicked the close button, sorry.

rv3r:
> Not fully fixed: if the app finds an empty slot before a slot with the bucket type, then it puts the items in the bucket slot, instead of stacking it.

Can I get before and after images of your inventory in this case? I need to see exactly what is going on in order to fix this.

Firigion:
Following screenshots demonstrate the two bugs I mentioned at once, because I can't separate them.

- Expected behaviour: buckets stack when right clicking a water sourche witha  bucket and I have a water bucket in my hotbar:
Before clicking
image
After clicking
image
**As you can see, buckets stack, but three were generated instead of one. This only happens when buckets are stacked, not for the first bucket in a stack, when vanilla behavour takes over.**

- "Buggy" behaviour: if there's an empty slot with a lower index that the first occupied slot, buckets go inot that empty slot, rather than stacking:
Before clicking
image
After clicking
image
Clicking again
image
**As you can see, first bucket goes into first empty slot, second bucket does stack as expected, but becomes three buckets, instead of one**

- Extra showcase of the duplication (multiplication) bug using lava:
Before click
image
After click
image

rv3r:
> "Buggy" behaviour: if there's an empty slot with a lower index that the first occupied slot, buckets go inot that empty slot, rather than stacking:

Should be fixed by 13cf295. I forgot that items fill *any* incomplete stack before finding the first empty slot.

> As you can see, buckets stack, but three were generated instead of one. This only happens when buckets are stacked, not for the first bucket in a stack, when vanilla behavour takes over.

Should be fixed by 4c8f9d9. Turns out the calculation for the number of items to move was poorly written.

Firigion (approved):
Awsome! Sorry for the nitpick, I wanted to make sure it was bug free :)

rv3r:
Thanks for testing it. Here's hoping it's finally done and can be added soon.

PR #213, opened 2021-06-01
Add new scripts to survival category

Crec0:
item_holding_mobs_despawn.sc 
> Mobs that hold items will despawn unless given a custom name using name tags. Does not change the way villager mobswitch works. More testing required. 

show_bee_count.sc
> prints the bee count in targeted beehive/bee nest block when right clicked on it with empty hand while crouching. Sends the count only to the player using the script. 

what_biome_on_block.sc
> prints out the biome name on blocks in a certain radius around the player. Useful for marking out biomes like river/ocean/soul sand valley above the roof and so on.


programs/survival/item_holding_mobs_despawn_rule.sc
```diff
@@ -0,0 +1,24 @@
...
+            modify(_, 'persistence', false);
+        )
+    )
+);
+
+global_counter = 0;
+
+__on_tick()->(
+    if (global_counter == 0,  schedule(0, '__change_persistence'));
+    global_counter = (global_counter + 1) % 20;
```
BisUmTo, line 23: You can use `if(! tick_time() % 20, ...)` instead

Crec0: Thanks for the suggestion. I will update it

programs/survival/item_holding_mobs_despawn_rule.sc
```diff
@@ -0,0 +1,21 @@
...
+    for ( entity_list('monster'),
+        if (_ ~ 'persistence' && _ ~ 'holds' && ! _ ~ 'custom_name',
+            modify(_, 'persistence', false);
+        )
+    )
+);
+
+__on_tick()->(
+    if (! tick_time() % 20,  schedule(0, '__change_persistence'));
+);
```
Firigion, line 21: I think it would be best to bind this to `entity_load_handler`, rather than a global `__on_tick` call. That wasy you don't need to check any mob more than once (and you also don't use the `__on_tick` method, which is disencouraged).

programs/survival/item_holding_mobs_despawn_rule.sc
```diff
@@ -0,0 +1,21 @@
+// changes despawn rule for monster mobs. 
+// makes mobs who hold items despawnable as opposed to vanilla where they will never despawn.
+// mobs can only be made despawnable when they are given a custom name using nametags
+
+__command()->null;
```
Firigion, line 5: Did you want to add a way to manually trigger the command and forgot?

programs/survival/show_bees_count.sc
```diff
@@ -0,0 +1,13 @@
+// shows the count of bees on that specific block
+// right click while sneaking on a beehive or beenest
+
+__command()->null;
```
Firigion, line 4: You only need to add `__command()` if you want the app to actually have a command tree. If it doesn't use player commands, you can just skip it.

programs/survival/what_biome_on_block.sc
```diff
@@ -0,0 +1,79 @@
+// shows biomes blockwise. useful for marking out rivers, beaches, soul sand valley and stuff.
+// now works across diemsions (yes even in the_end)
+__config()->{
+    'stay_loaded'->true,
+    'scope' -> 'global',
+    'commands' -> {
+        '' -> _() -> print(format('r Missing arguments. Correct syntax /showbiomes <toggle>')),
+        '<toggle>' -> l('__toggle_markings'),
```
Firigion, line 8: You don't need to input a list if you are not also inputting default values for missing arguments. Also, you can use `[·]` intead of `l(·)`.

programs/survival/what_biome_on_block.sc
```diff
@@ -0,0 +1,79 @@
...
+// if the script crashes your multiplayer world, do one of the two
+// 1. update carpet to latest version
+// 2. turn this to true 
+global_carpet_crash_fix = false;
+
+// change this global_radius to improve fps if you on a potato. 4 works decently
+global_radius = 4;
+
+// change the font size to your liking
+global_font_size = 6;
```
Firigion, line 25: If these are messages to the user, I'd move it to the top :)

programs/survival/what_biome_on_block.sc
```diff
@@ -0,0 +1,79 @@
...
+    l(posx, posy, posz)  = pos(player);
+    scan(posx, posy, posz, -global_radius, 0, global_radius,
+        l(x, y, z) = pos(_);
+        draw_shape(
+            'box', 
+            2,
+            'color', 0x00000000,
+            'fill', if(global_carpet_crash_fix, 0x808080FF, biome(_, global_players:player:'dim_color')) - 0x0000009F, 
+            'from', l(x, y + 0.05, z), 
+            'to', l(x + 1, y - 0.05, z + 1),
```
Firigion, line 41: Make hieght also configurable? Maybe add in-game configs?

programs/survival/what_biome_on_block.sc
```diff
@@ -0,0 +1,79 @@
+// shows biomes blockwise. useful for marking out rivers, beaches, soul sand valley and stuff.
+// now works across diemsions (yes even in the_end)
+__config()->{
+    'stay_loaded'->true,
+    'scope' -> 'global',
+    'commands' -> {
+        '' -> _() -> print(format('r Missing arguments. Correct syntax /showbiomes <toggle>')),
```
Firigion, line 7: Right now this is not true, the name of the app is "what_biome_on_block", hence the base command is also that. You can use `system_info('app_name')` to get the app's name to fill here.

programs/survival/what_biome_on_block.sc
```diff
@@ -0,0 +1,79 @@
...
+
+__toggle_markings(toggle) -> (
+    p = player();
+    if(!global_players:p, __add_player(p));
+    global_players:p:'toggle' = toggle == 'on';
+);
+
+__on_player_connects(player) -> (
+    __add_player(player);
+);
```
Firigion, line 67: If you don't make the app global scoped, you can remove this. Only reason I can see to make it global is that `print` shows the current setting for all online players, instead of just the caller? idk if that's intended.

Firigion (commented):
Nice! I really like the biome show app. Default radius is  a bit conservative, but it's okay :P

programs/survival/item_holding_mobs_despawn_rule.sc
Crec0: Entity load handler wont work in the case of when a mob picks up item after spawning. For example a zombie picking up block you left behind while mining. or does it ?

programs/survival/item_holding_mobs_despawn_rule.sc
Crec0: I wanted to disable the command. but that failed. Same for the bottom case. I just want the user to load the app and the command to not appear in the command list.

Crec0: Is there a way I can disable it ?

programs/survival/show_bees_count.sc
Crec0: Same as above

programs/survival/what_biome_on_block.sc
Crec0: gotcha. I once read that scarpet has to convert all the [] into l() and {} in to m(). is that not the case anymore ?

programs/survival/what_biome_on_block.sc
Crec0: ah right. I missed it. Showbiome was the original name for the app. But, it was not descriptive enough imo. Therefore I changed the name to be more descriptive and missed changing it here

programs/survival/what_biome_on_block.sc
Crec0: Print was the debug command left behind. I will update it.

Crec0: Also, is there a upside/downside of loading as global ? If there's none, I can for sure get rid of the global list.

programs/survival/what_biome_on_block.sc
Crec0: height relative to the player (y value) or height of the box drawn ?

programs/survival/item_holding_mobs_despawn_rule.sc
altrisi: Just don't write the line for it, it won't be registered then.

programs/survival/what_biome_on_block.sc
altrisi: It does, at compile time. There's no reason to remove readability by doing it manually.

programs/survival/what_biome_on_block.sc
altrisi: Loading app as global means there is a single instance of the app, sharing the same variables and everything else.

Player scope means there is an instance for every player, each with their own globals (and handlers I think?).

The overhead of one or the other is not noticeable, if any.

programs/survival/item_holding_mobs_despawn_rule.sc
altrisi: I think what Firi meant was to attach an `on_tick` entity event to the entity in load handler so there's no need to query the world and every entity every tick, and you are then just checking data in the selected entities.

programs/survival/what_biome_on_block.sc
Crec0: Isn't a single instance shared better than individual instances for all players ? Also if theres no upside of not using player scope, I will stick with global scope.

programs/survival/item_holding_mobs_despawn_rule.sc
Crec0: I have tried without the __command() line but it still shows up. :/

programs/survival/item_holding_mobs_despawn_rule.sc
Crec0: I will look into the documentation.

Crec0:
> Nice! I really like the biome show app. Default radius is  a bit conservative, but it's okay :P

What would be a better option for radius ? Ask the user in-game or ?

programs/survival/what_biome_on_block.sc
Firigion: I meant relative to the player, so thye can decide how far down form them the boxes are. Just a thought, tho. I usually lean towards more configurable apps that also work out of the box.

programs/survival/what_biome_on_block.sc
Firigion: I like show biomes, but my point was, since the base command is decided by the app's (file) name, any user can change it to whatever they want. To accomodate for that, `system_info('app_name')` returns the current app name as a string so you can use it in messages like this one.

programs/survival/what_biome_on_block.sc
Firigion: The upside is that variable handling is easier, since you don't ahve to keep a map indexing your settings for each player. Just a `global_state` will suffice, no need to query current player. Gnembon also sometimes says they are "safer", I think he referes to the fact that you don't get bugs from one player changing something for another player where the programmer didn't forsee it being possible.

programs/survival/item_holding_mobs_despawn_rule.sc
Firigion: Yes, that's what I meant.` entity_load_handler` also triggers when the entities are laoded in the world, not just newly spawned, iirc, so it should work fine. Don't quote me on this, but hooking up to the `on_tick` event of the mob is less expensive than going through the list of all entities, since the game doesn't need to query the entities. I think, not sure.

Firigion:
Meh, the radius thing was just a comment. Since the player can change it from in-game, its not a big issue. Personally, I set it to 10 when testing it, but if you leave it at four, no one will complain.

programs/survival/what_biome_on_block.sc
Firigion: This way you don't need to concern yourself with a `global_players` map and the corresponding `__on_player_connects` and `-disconnects` events and stuff like that.

programs/survival/showbiome.sc
```diff
@@ -0,0 +1,71 @@
...
+            'pos', [x + 0.5, y, z + 0.5],
+            'size', 6
+        );
+    );
+);
+
+__add_player(player) -> (
+    global_radius = 4;
+    global_player = {'dim_color' -> global_dimension_map:(player ~ 'dimension'), 'toggle' -> false};
+);
```
Firigion, line 54: If scope is `global`, global variables will be shared among playres, so setting `global_radius = 4` will effectively overwrite both the user modified (if any) `global_radius` in line 9 and the supposedly player-specific `global_radius` set with the `/showbiomes radius` command. If you are gonna keep the global scope, you need to use `global_radius` as a `global_radius_default` and then set for each player a `radius` inside the map, which `__make_markers` should use.

programs/survival/showbiome.sc
```diff
@@ -0,0 +1,71 @@
...
+    'arguments' -> {
+        'toggle' -> {'type' -> 'text', 'options' -> ['on', 'off']},
+        'radius' -> {'type' -> 'int', 'min' -> 1, 'max' -> 8, 'suggest' -> [2, 4, 8]}
+    }
+};
+
+global_player = {};
+global_dimension_map = {'overworld' -> 'foliage_color', 'the_nether' -> 'fog_color', 'the_end' -> 'sky_color'};
+
+__make_markers() -> (
```
Firigion, line 27: I missed this last time, but this doesn't seem to be multiplayer safe? not sure, since I didn't test, but `player()` will grab any player in the server when loading this app in `global_scope`, so what you'd need to do is iterate over the player list and run this funciton for each player (with toggle on).

programs/survival/showbiome.sc
```diff
@@ -0,0 +1,71 @@
...
+    delete(global_player);
+);
+
+__on_player_changes_dimension(player, from_pos, from_dimension, to_pos, to_dimension) -> (
+    if (global_player, global_player:'dim_color' = global_dimension_map:to_dimension);
+);
+
+__on_tick() -> (
+    if (global_player:'toggle', schedule(0, '__make_markers'));
+);
```
Firigion, line 71: Apparently I forgot to mention this the last time: since you use a command to first trigger the rendering of the biomes, why not use that same command to schedule the first `__make_markers` and make that schedule itself? That way you can forgo the `__on_tick` call completly and move to player-scope.

Firigion (commented):
More things on global vs player scope :P

programs/survival/showbiome.sc
Crec0: I have tested it in multiplayer and so far it only grabs the player who invokes the script. Because according to documentation player() does that

programs/survival/showbiome.sc
Crec0: Hmm thats sounds like a good idea. I will try it.

programs/survival/showbiome.sc
Crec0: I changed to scope to player in the last commit so it should be fine

Crec0:
> Meh, the radius thing was just a comment. Since the player can change it from in-game, its not a big issue. Personally, I set it to 10 when testing it, but if you leave it at four, no one will complain.

Anything above 8 starts to have rendering issues :p

Firigion (approved):
Awsome!

BisUmTo (approved):

altrisi:
Just noticed, why is the despawn with items app suffixed `rule`?

Crec0:
Maybe because it's changing an intended mechanic/despawning condition (or a rule) and that stuck with it? I am not sure.

PR #215, opened 2021-06-07
Add some matrix and vector maths

Firigion:
I added a few funcitons, not sure if all of them should stay, but this is why we have reviews.

Comments on functions:
* `norm`
* `normalize`
added because it's slightly better than doing `distance(vec, 0)` and `vec/distance(vec, 0)`, respectively.

* `cross`(_product)
* `outer`(_product)
could be renamed to *`_product`. I left like this to inkeep with current naming of dot prouct being `dot`.

* `cross_matrix`
could be removed (along with `outer`). I added them because they are needed (useful?) when doing rottion about an arbitrary axis in 3D, which is sometimes useful. I'd be totally okay with removing these.

* `mat_times_vec`
For 3D rotations about cartesian axes and some other geometric mumbo jumbo.

* `mat_times_mat`
* `transpose`
for completeness. I haven't had to use them so far in any app. Again, could be removed. `mat_times_mat` is not particulrly efficient either, but I couldn't find a neat vectorized way to do it like I did for `mat_times_vec`.

* `direction`
There has been a bunch of people doing the maths themselves and reinventing the wheel. Why not provide users with the function?

### DIFF
--- a/programs/fundamentals/math.sc
+++ b/programs/fundamentals/math.sc
@@ -42,6 +42,35 @@ distance(vec1, vec2) -> sqrt(reduce(vec1 - vec2, _a + _*_, 0));
 
 dot(vec1, vec2) -> reduce(vec1 * vec2, _a + _, 0);
 
+norm(vec) -> sqrt(reduce(vec, _a + _*_, 0));
+
+normalize(vec) -> vec / sqrt(reduce(vec, _a + _*_, 0));
+
+cross(v, w) -> [v:1*w:2 - v:2*w:1, v:2*w:0 - v:0*w:2, v:0*w:1 - v:1*w:0];
+
+outer_prod(v, w) -> map(w, v*_);
+
+corss_matrix(vec) -> (
+	[x, y, z] = vec;
+	[[0, -z, y], [z, 0, -x], [-y, x, 0]]
+);
+
+// the following methods assume a square matrix indexed by row (i.e, mat:0 is the first row)
+mat_times_vec(matrix, vec) -> map(matrix * map(vec, vec), reduce(_, _a+_, 0));
+
+mat_times_mat(mat1, mat2) -> (
+	mat2 = transpose(mat2);
+	map(mat1,
+		i = _i;
+		map(mat2,
+			j = _i;
+			dot(mat1:i, mat2:j)
+		)
+	)
+);
+
+transpose(mat) -> map(mat, i = _i; map(_, mat:_i:i) );
+
 highest_common_factor(num1,num2) -> (
 	q=1;
 	while(q!=0,num1*num2,
@@ -61,3 +90,6 @@ highest_common_factor_list(list)-> reduce(list, highest_common_factor(_,_a),list
 
 // computes LCM for a list of numbers
 lowest_common_multiple_list(list)-> reduce(list, lowest_common_multiple(_,_a), list:0);
+
+// convert yaw, pitch as given by MC into a normalized vector
+direction(yaw, pitch) -> [-sin(yaw)*cos(pitch), -sin(pitch), cos(pitch)*cos(yaw)];


Ghoulboy78:
Yo Firi, I made a whole scarpet script for that. It's high time I committed it, but it basically contains all this (minus the vector stuff), implemented in a safe and idiot proof way, so it throws appropriate errors if the user didn't input a list or matrix, as opposed to giving a garbled return value or some other parsing error which would be difficult to understand.

Firigion:
Ah, that's nice. I thought about doing that, but since none of the `math` functions are protected, I decided to keep the same concept.

Firigion:
@Ghoulboy78 since there's no ETA for your PR, would you mind approving this one until you work on yourse? Doesn't make sense to block a PR because at some point in the probable future a better one will be merged. Ca jsut delete the additions form this one once yours is done.

programs/fundamentals/math.sc
```diff
@@ -42,6 +42,35 @@ distance(vec1, vec2) -> sqrt(reduce(vec1 - vec2, _a + _*_, 0));
 
 dot(vec1, vec2) -> reduce(vec1 * vec2, _a + _, 0);
 
+norm(vec) -> sqrt(reduce(vec, _a + _*_, 0));
+
+normalize(vec) -> vec / sqrt(reduce(vec, _a + _*_, 0));
+
+cross(v, w) -> [v:1*w:2 - v:2*w:1, v:2*w:0 - v:0*w:2, v:0*w:1 - v:1*w:0];
```
Ghoulboy78, line 49: is there a more generalized algorithm for the cross product of vectors? cos in that case I would prefer to use that, but this is fine till then.

programs/fundamentals/math.sc
```diff
@@ -42,6 +42,35 @@ distance(vec1, vec2) -> sqrt(reduce(vec1 - vec2, _a + _*_, 0));
...
+
+outer_prod(v, w) -> map(w, v*_);
+
+corss_matrix(vec) -> (
```
Ghoulboy78, line 53: corss...

Ghoulboy78 (approved):
Ye, this looks good, I'll nick stuff from here when I get around to finishing my matrix pr.

programs/fundamentals/math.sc
Firigion: what do you mean? Cross product is only an operation that exists in R³ between vectors. You can think of it in terms of the determinant of a matrix where the first row is the versors of the three directions and the second and third are the multiplied vetors, but that wouldn't be any faster than this.

Ghoulboy78: ah k

Ghoulboy78: Wait, apparently you can make cross product for vectors of length>3, but it's a pain, and I also just realised ur only doing vectors of length 3 here so nvm

altrisi:
I'd add matrix and vector math to its own library, instead of the general math one (that shouldn't exist here with the same `math` name, but that's another issue). Also corss was never changed.

programs/fundamentals/math.sc
```diff
@@ -42,6 +44,35 @@ distance(vec1, vec2) -> sqrt(reduce(vec1 - vec2, _a + _*_, 0));
...
+	map(mat1,
+		i = _i;
+		map(mat2,
+			j = _i;
+			dot(mat1:i, mat2:j)
+		)
+	)
+);
+
+transpose(mat) -> map(mat, i = _i; map(_, mat:_i:i) );
```
Ghoulboy78, line 74: Umm, this doesn't work on non-square matrices... `transpose([[1, 2], [3, 4], [5, 6]])` should return `[[1, 3, 5], [2, 4, 6]]`, instead it returns: `[[1, 3], [2, 4], [1, 3]]`

Firigion:
> Also corss was never changed.
Ohhh _that's_ what the comment meant. I literally don't notice that sort of typo. Dyslexia and all, you know.

programs/fundamentals/math.sc
Firigion: True, true. Should add a disclaimer? I'm not gonna fix it, but we can delete it, if you don't like the disclaimer. People are ognna open the app to know what operations are available anyway.

Firigion:
Regarding name, shouldn't _this one_ be the default maths library? The built-in one never gets updated and stuff here is never gonna get used if the library is not packeged with the app, since `/script download` doesn't download libraries (I think?)

programs/fundamentals/math.sc
FR33TR1ST: I have an idea to how to fix it, but i dont know how to program scarpet, i can do a fix using python structure if it usefull

Firigion: Yeah, post python code here, i'll import it to scarpet.

PR #216, opened 2021-06-07
Fixed a few bugs, implemented  afew features

Firigion:
This app's development is being tracked in https://github.com/boyenn/scarpet-scripts. Changelog would read something like:

* Configurable track lines (thickness, angle, length)
* Bugfix: Appropiate dimension support.

### DIFF
--- a/programs/survival/waypoints.sc
+++ b/programs/survival/waypoints.sc
@@ -7,7 +7,7 @@ global_waypoint_config = {
     // 4 : ALWAYS
     'allow_tp' -> 2
 };
-print(player());
+
 _can_player_tp() -> (
     global_waypoint_config:'allow_tp' == 4 ||
     ( global_waypoint_config:'allow_tp' == 3 && player()~'permission_level' > 1) || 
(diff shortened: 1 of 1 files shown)


Firigion:
Btw, this is being used in Hammer and Melon SMPs, as well as some other smaller private ones, so it has quite some field testing. I only posted the update after dispatching it in those servers and having the players use them for a bit.

programs/survival/waypoints.sc
```diff
@@ -7,7 +7,7 @@ global_waypoint_config = {
     // 4 : ALWAYS
     'allow_tp' -> 2
 };
-
+print(player());
```
altrisi, line 10: Assuming this is debug code that shouldn't be there?

altrisi (approved):
Other than that line LGTM (didn't launch it though).

programs/survival/waypoints.sc
Firigion: Indeed. I totally missed it because, since it's a global app, the essage is only printe on server startup, so I never saw it printing.

BisUmTo (approved):

PR #217, opened 2021-06-08
Adding matrix.sc

Ghoulboy78:
This adds a bunch of functions which are related to matrices, including ones to create new matrices, check that a value is a matrix, and multiply, add and subtract. Inverse matrices are (obviosly) kinda more difficult to do, so I've left them out for the time being.

IMO some of the functions which check if a number is a matrix or a vector could be moved elsewhere, maybe to math.scl

Also, should I change the name to matrix.scl?

### DIFF
--- a/programs/fundamentals/matrix.sc
+++ b/programs/fundamentals/matrix.sc
@@ -1,0 +1,178 @@
+//A library of functions to create and manipulate matrices. These are as of now just lists of lists of numbers, but with these functions can act as matrices
+//This is of course an abstract library, which will no doubt be replaces by a better native scarpet implementation in the future, but it will do for now.
+//This may be temporary in case gnembon adds matrices into scarpet as a built in data structure, but nothing is certain so far.
+//By: Ghoulboy
+
+sum(...args)->reduce(args, _a+_, 0);//todo add to maths library cos for some reason it's not a thing yet?
+
+create_matrix(width, height)->(//creates an empty matrix (so all 0 values)
+    matrix = [];
+    loop(height,
+        matrix: _= [];
+        w = _;
+        loop(width,
+            matrix:w:_ = 0
+        );
+    );
+    matrix
+);
+
+_create_matrix(...columns)->( //creating matrix a bunch of columns, or as a singleton list of columns
+    if(is_matrix(columns),
+        matrix = columns,
+        check_matrix(columns:0); //if this throws an error (i.e that is not a matrix) then it will fail, if not return null and continue as normal
+        matrix = columns:0
+    );
+    matrix
+);
+
+is_matrix(matrix)->(//returns boolean, true if given value is a matrix and false otherwise
+
+    try(//cos scarpet try doesn't actually catch errors, only when you call throw() function
+
+        if(type(matrix) == 'list' && length(matrix)>0 && type(matrix:0)=='list',//doing easy checks first, so it fails early without having to do excess computation.
+
+            //checking if all columns are the same length
+            col_size = length(matrix:0);
+            if(col_size==0, throw());//checking that it's not got empty columns (cos that technically passes the next check)
+
+            //checking if all the items are numbers
+            loop(length(matrix),
+                row = _;
+
+                if(length(matrix:_)!=col_size, throw()); //checking that all columns are the same length
+
+                loop(col_size,
+                    if(type(matrix:row:_)!='number',
+                        throw()
+                    )
+                )
+            ),
+
+            throw()
+        ),
+        return(false)// todo decide whether or not to give more detailed error, i.e what exactly went wrong
+    );
+    true
+);
+
+is_vector(vector)->//checking like this so the first check fails without having to iterate over the whole list potentially
+    !(type(vector)!='list' || length(vector)==0 || first(vector, type(_)!='number'));
+
+check_matrix(matrix)->( //checks if given value is a matrix, and if not throws an error
+    if(!is_matrix(matrix),
+        exit(print('Operation is invalid, "' + matrix + '" is not a matrix'))
+    )
+);
+
+
+identity(size)->(//returns identity matrix of given size
+    identity = [];
+    loop(size,
+        width = _;
+        identity:_=[];
+        loop(size, //cos its always a square matrix
+            identity:width:_ = width ==_
+        )
+    );
+    identity
+);
+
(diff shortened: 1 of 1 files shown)


altrisi:
Didn't fully review it, but I'd throw custom exceptions like `invalid_matrix` maybe? (so lib users can filter them). 
And with a bit of info of why it failed.

I'd make it a library given it doesn't hold variables by itself.

Also don't `exit()`, that fully finishes script execution in that stack. Apps may want to catch that. Also don't print messages if it's going to be a library, apps may catch those errors and then it would be weird to a player.

programs/fundamentals/matrix.sc
```diff
@@ -0,0 +1,176 @@
...
+
+matrix_subtract(a, b) ->( //returns a new matrix whose values are the difference of the two input matrices (i.e a - b)
+    check_matrix(a);
+    check_matrix(b);
+
+    if(length(a)!=length(b) || length(a:0)!=length(b:0),
+        exit(print('Cannot subtract uneven matrices'))
+    );
+
+    sum = [];
```
Firigion, line 119: `dif` or something like that

programs/fundamentals/matrix.sc
```diff
@@ -0,0 +1,176 @@
...
+
+    loop(length(a),
+        width = _;
+        sum:width=[];
+        loop(length(a:0),
+            sum:width:_ = a:width:_ - b:width:_
+        )
+    );
+    sum
```
Firigion, line 128: same

programs/fundamentals/matrix.sc
```diff
@@ -0,0 +1,176 @@
+//A library of functions to create and manipulate matrices. These are as of now just lists of lists of numbers, but with these functions can act as matrices
+//This is of course an abstract library, which will no doubt be replaces by a better native scarpet implementation in the future, but it will do for now.
+//This may be temporary in case gnembon adds matrices into scarpet as a built in data structure, but nothing is certain so far.
+//By: Ghoulboy
+
+create_matrix(width, height)->(//creates an empty matrix (so all 0 values)
```
Firigion, line 6: A function like this one is usually called `zeros` or something along those lines. If you like, call it `zero_matrix`, but I'd suggest explicitly stating the fact that it generates a matrix of zeros in its name.

programs/fundamentals/matrix.sc
```diff
@@ -0,0 +1,176 @@
...
+    matrix = [];
```
Firigion, line 7: In general for the whole app/library: isn't `map` more efficient than generating an empty lsit and then expanding it at every step?

programs/fundamentals/matrix.sc
```diff
@@ -0,0 +1,176 @@
...
+    );
+    sum
+);
+
+//This is complicated, because you can do many things here, depending if input is scalar, vector or matrix:
+//  scalar * scalar = error (y doin that op here)
+//  scalar * vector = error (y doin that op here)
+//  scalar * matrix = error (cannot do it in this order). Still gotta handle.
+//  vector * scalar = error (y doin that op here)
+//  vector * vector = matrix with rows same as length of first vector, columns same as that of second.
```
Firigion, line 136: This is called outer product of vectors, it's what happens when you multiply two "matrices" of sizes N×1 and 1×N, respectivelye (while the inverse 1×N times N×1 results in the usual dot product). If your "vector" objects are just gonna be lists, I'm not sure it's a good idea to include this behaviour in this function.

Put in another way, you need to decide what shape your vectors represent: is a list a 1×N (row vector) or a N×1 (column vector) object? In the latter case, outer product is `v×w transposed`, and `v transposed × w` is dot product.

programs/fundamentals/matrix.sc
```diff
@@ -0,0 +1,176 @@
...
+                        result:row = reduce(b:row, _a + (_ * a:row),0);
+                    ),
+                    exit(print(str('Cannot multiply vector %s by matrix %s in this order, they have mismatched sizes', a, b)))
+                ),
+                exit(print('Why are you doing that operation here?'))
+            ),//in case they inputted a list which wasn't a vector.
+            is_matrix(a), //matrix * something
+            if(is_vector(b), //matrix * vector = vector
+                
+            ),
```
Firigion, line 167: mat×mat and mat×vec yet to be implemented?

Firigion (commented):
I like where this is headed. I think that if you are gonna include vectors, it's a good idea to decide on if they represent a 1×N or a N×1 matrix by default.

Also, I think separating multiplication of vectors and matrices in their different combinations into separate functions might be a good idea, for speed's sake. At least, have one for outer product, one for matrix product and one for mat×vec and vec×mat.

Firigion:
Yeah, I think it should be `.scl`. Also, I think we should check if the vectorized form of the operatios is faster than the element wise with the two loops. If that's the case, there's vectorized versions for almost all operations, but a custom one for each one at that. See #215 for my implementations, which I don't claim to be fast, but I think they are worth a test.

programs/fundamentals/matrix.sc
Ghoulboy78: Yes, I hadn't gotten round to it cos I kinda forogt about this until I saw ur other pr

programs/fundamentals/matrix.sc
Ghoulboy78: I think it makes more sense for vectors to be treated as 1*N matrices in this case, cos of the way they look in code: `[a, b, c]`, which looks more like a row vector than a column vector imo.

Firigion: I agree, although it's customary for vectors to default to N×1 matrices. Either way, you need ot make a choice, make it explicit and stick with it, imo.

Ghoulboy78:
Marking this as draft cos Im kinda busy these days, so I want some time to work on this to make it rly good.

programs/fundamentals/matrix.sc
Ghoulboy78: k

Ghoulboy78:
I'm putting this on hold until the fate of #306 is decided, given that if it's accepted then I'm going to rewrite both vector.scl and matrix.scl as classes, This is because no one's going to use them in a practical sense anyways, so they might as well get some extra syntactic sugar to make their code look prettier if they do use them.

Ghoulboy78:
Ok I've moved `matrix.scl` to use `classes.sc` and be a class. I will shortly do the same with `vector.scl`

PR #220, opened 2021-06-10
add lava_sponge.sc

B-2U:


programs/survival/lava_sponge.sc
```diff
@@ -0,0 +1,39 @@
...
+   m(
+      l('stay_loaded','true')
+   )
+);
+
+
+__on_player_places_block(player, item_tuple, hand, block)->(
+    if(hand != 'mainhand' || item_tuple:0 != 'sponge' || all(neighbours(pos(block)),_!='lava'), return());
+    set(pos(block),'wet_sponge');
+    global_count = 0;
```
Firigion, line 15: Why have them as global, instead of passing them to the funciton?

Firigion (commented):
I like how you removed the manhattan distance requirement in favour of a recursion count. My only comment is that you can make this slightly more efficient by remembering which blocks you already checked, because checking a ser with `has()` is way faster than checking the world for the block (with `_ == 'lava' `), so you could save a few checks there.

Also, in general `[·]` and `{·}` are more readable than `l(·)` and `m(·)`, but it's not "wrong" the way you did it.

Firigion:
Add the script to the folder's readme, plz

programs/survival/lava_sponge.sc
```diff
@@ -0,0 +1,39 @@
...
+            if(_ == 'lava' && global_recursion<=7 && global_count<=64,
+                now_pos = pos(_);
+                if(all(lava, _!=now_pos),
+                    put(lava, null, pos(_));
+                    global_count +=1;
+                )
+            )      
+        )
+    );     
+    if(global_lava!=[] && global_recursion<=7 && global_count<=65,
```
Firigion: Woops, I missed this, and so did you. It's no longer global.

B-2U: sloved in add5aa8

programs/survival/lava_sponge.sc
```diff
@@ -0,0 +1,43 @@
...
+        for(neighbours(coord), 
+            if(_ == 'lava' && recursion<=7 && count<=64,
+                if(!has(lava, pos(_)),
+                    lava += pos(_);
+                    count +=1;
+                )
+            )      
+        )
+    );     
+    if(lava!={} && recursion<=7 && count<=65,
```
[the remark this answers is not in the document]
B-2U: i tried to run some script like
 `/script run a={}; c_for(d=0,d<1000,d+=1,a+=d); if(a!={}, print('false'))`
and
 `/script run a={}; c_for(d=0,d<1000,d+=1,a+=d); if(length(a)!=0, print('false'))`
their runtime are basically the same
not sure if this script can show the difference though

B-2U (commented):
put the block type check to the last of four condition

programs/survival/lava_sponge.sc
```diff
@@ -0,0 +1,43 @@
...
+    recursion = 0;
+    sponge(l(pos(block)), recursion, count);
+);
+
+sponge(coord_list, recursion, count)->(
+    lava = {};
+    for(coord_list,
+        coord = _;
+        for(neighbours(coord), 
+            if(_ == 'lava' && recursion<=7 && count<=64,
```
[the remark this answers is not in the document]
B-2U: optimize in #84ff435

programs/survival/lava_sponge.sc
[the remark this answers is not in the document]
Ghoulboy78: well for me the first one took 38ms, and the second one took 0.4ms...
To compare them both, it would be best to have a function which adds say 1000 items to `a`, and then at each step do `profile_expr(a=={})` and `profile_expr(length(a)==0)`, and then see how they increase over time. Though off the top of my head I can tell you that a=={} will be O(n) cos you have to check each item, whereas length(a)==0 is designed to be O(1).

programs/survival/lava_sponge.sc
```diff
@@ -0,0 +1,43 @@
...
+    recursion = 0;
+    sponge(l(pos(block)), recursion, count);
+);
+
+sponge(coord_list, recursion, count)->(
+    lava = {};
+    for(coord_list,
+        coord = _;
+        for(neighbours(coord), 
+            if(!has(lava, pos(_)) && recursion<=7 && count<=64,
```
Ghoulboy78, line 25: could separate the check for recursion to be at the top of the function to avoid repetive checks, and the check for count to be in an if statement outside this one, so that if it fails, you can break out of the loop, and save some compute time in looping needlessly and checking the same condition over and over.

programs/survival/lava_sponge.sc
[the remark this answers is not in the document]
B-2U: I see, Ill use length instead

programs/survival/lava_sponge.sc
B-2U: Do you mean a structure like `if recursion( if count ( if has(lava) ( if block == 'lava')))) ` ?

Ghoulboy78: no. I mean `if(count etc, if(lava), break())`, and put the recursion check outside the loop cos the recursion variable is updated outside the loop, so checking it over and over again is effectively a waste of time. Generally with recursion you check the recursive depth at the top of the function.

B-2U: My idea was recursion and count are just simple int check, so i put them together to get prettier code look

B-2U: plz check the latest commit

programs/survival/lava_sponge.sc
```diff
@@ -0,0 +1,44 @@
...
+__config() -> (
+   m(
+      l('stay_loaded','true')
+   )
+);
+
+
+__on_player_places_block(player, item_tuple, hand, block)->(
+    if(hand != 'mainhand' || item_tuple:0 != 'sponge' || all(neighbours(pos(block)),_!='lava'), return());
+    set(pos(block),'wet_sponge');
```
Ghoulboy78, line 14: You can avoid the return call, and instead wrap the code in an if statement:
```
if(hand=='mainhand' && item_tuple:0=='sponge' && any(neighbours(pos(block)), _=='lava'),
    sponge
)
```
Thats just cos return, break and exit calls tend to be slower for scarpet parser, so it's best not to use them when possible

B-2U: k

Ghoulboy78 (approved):
Ok, this is good, Ill merge it

B-2U:
> Ok, this is good, Ill merge it

oh sry, new to github, seems i pressed something want you to review it again

Ghoulboy78 (approved):
nvm, turned out still needs another review to merge

Firigion (approved):
Awsome!

PR #221, opened 2021-06-12
Update Cam: Actionbar instead of chat, simplify and fix timeout

altrisi:
This PR updates the cam app to change a few things:

- Simplify timeout code:
  Previously, there were two branches for the switch code: one if there was timeout, another one if there wasn't. This PR simplifies it by instead of scheduling to timeout+1sec, scheduling to timeout, so if it's 0 it will run instantly (0*20 = 0). This also brings the next upgrade:
- Fix timeout not showing directly on command output and being +1:
  Until now timeout messages were sent since one second after the command was ran, which not only could make users wonder if it didn't work, but also made the timeout effectively one extra second.
- Remove empty chat messages:
  At some point, quite long ago, it was changed that in command functions, if they return an empty string they would send it to chat. Previously Scarpet would consider that "don't send anything", but it was changed and produced empty lines, so I changed that return to null.
- Use actionbar for messages:
  Makes use of the new actionbar API instead of chat, since it's something you really only need on that moment. Also adds an "entered/exited camera mode" message to inform the user (and to clear the entering in 1 message).
  Color was changed because it was already not really visible IMO, and since actionbar is smaller, it was even less visible
- Newer syntax and removal of unneeded code:
  Changes the config to use the new map syntax and removes a single-item to list converter in a nbt list getter, since that was already fixed in a PR to Carpet not that long ago.

Please give feedback! I'm also not sure about the entered/exited messages, and I'm even less sure about the color.

### DIFF
--- a/programs/survival/cam.sc
+++ b/programs/survival/cam.sc
@@ -10,8 +10,8 @@ global_survival_timeout = 3;
 // additional player checks are here. comment lines that you think are not needed
 __assert_player_can_cam_out(player) ->
 (
-   if(!query(player,'nbt','OnGround'), exit('You must be on firm ground.'));
-   if(query(player,'nbt','Air') < 300, exit('You must be in air, not suffocating nor in liquids.'));
+   if(!query(player, 'nbt', 'OnGround'), exit('You must be on firm ground.'));
+   if(query(player, 'nbt', 'Air') < 300, exit('You must be in air, not suffocating nor in liquids.'));
    if(player ~ 'is_burning', exit('You must not be on fire.'));
    null
 );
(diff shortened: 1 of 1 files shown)


Firigion:
Yes, I like the timer changes a lot. I'm not sure if entered/exited camera mode should be actionbar or chat messages, like when you succesfully run a command in vanilla, that you get a light or dark gray message in chat. Other than that, awsome.

Ghoulboy78:
I tested it and saw that perhaps the colour of the action bare message is a tad too light, cos I had a hard-ish time seeing it cos my testing world is at daytime and full brightness. Also, perhaps you could put an emphasis on the numbers in the countdown by like making them bold and perhaps another colour?

Firigion (approved):
After testing this, I do prefer action bar over chat message

Ghoulboy78 (approved):

PR #222, opened 2021-06-21
add silk touch budding amethyst script

DragonCrafted87:
add silk touch for budding amethyst script

also cleanup a bunch of whitespace in the survival readme apparently

### DIFF
--- a/programs/survival/silk_budding_amethyst.sc
+++ b/programs/survival/silk_budding_amethyst.sc
@@ -9,16 +9,11 @@ __config() -> {'stay_loaded' -> true, 'scope' -> 'global'};
 //  - player is holding a silk touch diamond or netherite pickaxe
 //  - block is a budding_amethyst
 __silk_budding_amethyst (player, block) -> (
-    holds = player ~ 'holds';
-    if (
-        holds,
-        (
-            [tool, count, nbt] = holds;
-            block=='budding_amethyst' &&
-            tool ~ ['diamond_pickaxe', 'netherite_pickaxe'] &&
-            nbt ~ 'silk_touch';
-        )
-    )
+    [tool, count, nbt] = player ~ 'holds' || ['None', 0, null];
+
+    block=='budding_amethyst' &&
+    tool ~ ['diamond_pickaxe', 'netherite_pickaxe'] &&
+    nbt ~ 'silk_touch';
 );
 
 
--- a/programs/survival/silk_budding_amethyst.sc
+++ b/programs/survival/silk_budding_amethyst.sc
@@ -9,12 +9,17 @@ __config() -> {'stay_loaded' -> true, 'scope' -> 'global'};
 //  - player is holding a silk touch diamond or netherite pickaxe
 //  - block is a budding_amethyst
 __silk_budding_amethyst (player, block) -> (
-    [tool, count, nbt] = player ~ 'holds';
-    block=='budding_amethyst' &&
-    tool ~ ['diamond_pickaxe', 'netherite_pickaxe'] &&
-    nbt ~ 'silk_touch';
+    holds = player ~ 'holds';
+    if (
+        holds,
+        (
+            [tool, count, nbt] = holds;
+            block=='budding_amethyst' &&
+            tool ~ ['diamond_pickaxe', 'netherite_pickaxe'] &&
+            nbt ~ 'silk_touch';
+        )
+    )
 );
-
 
 
 __on_player_breaks_block (player, block) -> (
--- a/programs/survival/silk_budding_amethyst.sc
+++ b/programs/survival/silk_budding_amethyst.sc
@@ -33,8 +33,11 @@ __silk_budding_amethyst (player, block) -> (
 );
 
 __on_player_breaks_block (player, block) -> (
-        if (!__silk_budding_amethyst(player, block), return());
-        data = block_data(pos(block));
-		item_nbt = nbt('{Item:{id:"minecraft:budding_amethyst",Count:1b,PickupDelay:10}}');
-		budding_amethyst_item = spawn('item', pos(block) + [0.5,0.5,0.5], item_nbt);
+    if (
+        __silk_budding_amethyst(player, block),
+        (
+            item_nbt = nbt('{Item:{id:"minecraft:budding_amethyst",Count:1b,PickupDelay:10}}');
+            budding_amethyst_item = spawn('item', pos(block) + [0.5,0.5,0.5], item_nbt);
+        ),
+        return());
 );
--- a/programs/survival/silk_budding_amethyst.sc
+++ b/programs/survival/silk_budding_amethyst.sc
@@ -18,11 +18,9 @@ __silk_budding_amethyst (player, block) -> (
 
 
 __on_player_breaks_block (player, block) -> (
-    if (
-        __silk_budding_amethyst(player, block),
-        (
-            item_nbt = nbt('{Item:{id:"minecraft:budding_amethyst",Count:1b,PickupDelay:10}}');
-            budding_amethyst_item = spawn('item', pos(block) + [0.5,0.5,0.5], item_nbt);
-        )
+    __silk_budding_amethyst(player, block) &&
+    (
+        item_nbt = nbt('{Item:{id:"minecraft:budding_amethyst",Count:1b,PickupDelay:10}}');
+        budding_amethyst_item = spawn('item', pos(block) + [0.5,0.5,0.5], item_nbt);
     )
 );
--- a/programs/survival/silk_budding_amethyst.sc
+++ b/programs/survival/silk_budding_amethyst.sc
@@ -1,0 +1,31 @@
+// Conversion of Silk Spwaners; Implemented by Scott Gudeman (DragonCrafted87)
+// See https://github.com/gnembon/scarpet/blob/master/programs/survival/silk_spawners.sc for Original Basis
+
+//keeps script loaded upon server start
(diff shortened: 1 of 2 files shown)


programs/survival/silk_budding_amethyst.sc
```diff
@@ -0,0 +1,31 @@
...
+        nbt = get(tool, 2);
+        if (!nbt, return());
+        ench = get(nbt, 'Enchantments[]');
+        if (!ench, return());
+        if (type(ench) != 'list', ench = l(ench));
+        for (ench, if (get(_, 'id') == 'minecraft:silk_touch', return(true)));
+        false
+);
+
+__on_player_clicks_block (player, block, face) -> (
```
Ghoulboy78, line 24: I would recommend doing the check only when a player breaks a block, so it happens less often, and the block is not broken instantly. I know that that's how it's done in silk spawner script, but that's cos spawners have nbt data which is more complex to handle. Here, you're just dropping the block, so you can wait for the block to actually break.

programs/survival/silk_budding_amethyst.sc
```diff
@@ -0,0 +1,31 @@
...
+        if (block != 'budding_amethyst', return());
+        tool = player ~ 'holds';
+        if ((!tool || (get(tool, 0) != 'diamond_pickaxe' && get(tool, 0) != 'netherite_pickaxe')), return());
+        nbt = get(tool, 2);
+        if (!nbt, return());
+        ench = get(nbt, 'Enchantments[]');
+        if (!ench, return());
+        if (type(ench) != 'list', ench = l(ench));
+        for (ench, if (get(_, 'id') == 'minecraft:silk_touch', return(true)));
+        false
```
Ghoulboy78, line 21: Try to avoid using unnecessary `return()` calls, instead put it all into an if statement, checking if it's a budding amethyst block, and if the player is holding a netherite or diamond pick with silk touch. If you get creative (not that you have to though), you could squeeze it all into a single line, though that may be slow and hard to read/understand

programs/survival/silk_budding_amethyst.sc
```diff
@@ -0,0 +1,31 @@
...
+);
+
+__on_player_clicks_block (player, block, face) -> (
+        if (!__silk_budding_amethyst(player, block), return());
+        data = block_data(pos(block));
```
Ghoulboy78, line 26: why need to save block data if you're not using it?

programs/survival/silk_budding_amethyst.sc
```diff
@@ -0,0 +1,31 @@
...
+		item_nbt = nbt('{Item:{id:"minecraft:budding_amethyst",Count:1b}}');
+		budding_amethyst_item = spawn('item', pos(block), item_nbt);
```
Ghoulboy78, line 28: Normally items are spawned in the centre of the block (`pos(block) + [0.5,0.5,0.5]`), and with a little 'plop' sound, which would be nice imo.

programs/survival/silk_budding_amethyst.sc
```diff
@@ -0,0 +1,31 @@
...
+		modify(budding_amethyst_item,'pickup_delay',10);
```
Ghoulboy78, line 29: Can't you add that to the nbt string? (`item_nbt = nbt('{Item:{id:"minecraft:budding_amethyst",Count:1b, PickupDelay:10}}');`)

Ghoulboy78 (commented):
It seems quite good, just have a few QOL suggestions to improve it just a notch

programs/survival/silk_budding_amethyst.sc
Ghoulboy78: The reason for fewer return calls is cos they're kinda slow for the scarpet interpreter, so doing them very often is slow.

DragonCrafted87:
> It seems quite good, just have a few QOL suggestions to improve it just a notch

thanks, this is the first time I've really started to look at implementing new stuff in scarpet so definitely way off from any sort of "best practices" or even a good handle on all the functionality 

I hope I can find time after work today to do the cleanup.

Ghoulboy78:
Well, that's great for a first time! (You should see my first scarpet app's original code, it's such a mess...)
And don't worry about time, we all know that no one has any time at all these days, take care of irl first always.

programs/survival/silk_budding_amethyst.sc
DragonCrafted87: done and now the block still takes the default break time

programs/survival/silk_budding_amethyst.sc
DragonCrafted87: wound up basically rewriting the whole function and was actually able to simplify it a bit with some of the examples i found in the api :)

programs/survival/silk_budding_amethyst.sc
DragonCrafted87: moved the spawn position, didn't need to add any sound as Minecraft seems to be playing that on its own at least from what I hear, and the subtitles tell me

programs/survival/silk_budding_amethyst.sc
```diff
@@ -0,0 +1,43 @@
...
+);
+
+__on_player_breaks_block (player, block) -> (
+    if (
+        __silk_budding_amethyst(player, block),
+        (
+            item_nbt = nbt('{Item:{id:"minecraft:budding_amethyst",Count:1b,PickupDelay:10}}');
+            budding_amethyst_item = spawn('item', pos(block) + [0.5,0.5,0.5], item_nbt);
+        ),
+        return());
```
Ghoulboy78, line 42: No need for return() call here, nothing will happen. Just do `if(__silk_budding_amethyst, etc etc)`, it will work fine

programs/survival/silk_budding_amethyst.sc
```diff
@@ -0,0 +1,43 @@
...
+                holds,
+                [tool, count, nbt] = holds;
+                if(
+                    (
+                        tool ~ 'diamond_pickaxe' ||
+                        tool ~ 'netherite_pickaxe'
+                    ) &&
+                    nbt ~ 'minecraft:silk_touch',
+                    (
+                        return(true);
```
Ghoulboy78, line 26: No need to put return here, just put `true` and it will work

programs/survival/silk_budding_amethyst.sc
```diff
@@ -0,0 +1,43 @@
...
+                    )
+                )
+            )
+        ),
+        return(false);
```
Ghoulboy78, line 31: Same here, just put `false`

Ghoulboy78 (commented):
Here, just some finishing touches to polish it up a little.
To be completely honest though the entire `__silk_budding_amethyst` function could be collapsed into:
```
__silk_budding_amethyst(player, block)->(
    block=='budding_amethyst' && ['diamond_pickaxe', 'netherite_pickaxe'] ~ player~'holds':0 && player~'holds`:2 ~ 'silk_touch'
);
```
With proper spacing, that could be shorter, tho ig it's a bit slower cos you query `player~'holds'` twice.

DragonCrafted87:
> ```
> __silk_budding_amethyst(player, block)->(
>     block=='budding_amethyst' && ['diamond_pickaxe', 'netherite_pickaxe'] ~ player~'holds':0 && player~'holds`:2 ~ 'silk_touch'
> );
> ```

I broke this up into 2 statements, so we aren't getting the data twice and broke the line at that boolean ops just so its more readable

programs/survival/silk_budding_amethyst.sc
```diff
@@ -0,0 +1,26 @@
...
+    tool ~ ['diamond_pickaxe', 'netherite_pickaxe'] &&
+    nbt ~ 'silk_touch';
+);
+
+
+__on_player_breaks_block (player, block) -> (
+    __silk_budding_amethyst(player, block) &&
+    (
+        item_nbt = nbt('{Item:{id:"minecraft:budding_amethyst",Count:1b,PickupDelay:10}}');
+        budding_amethyst_item = spawn('item', pos(block) + [0.5,0.5,0.5], item_nbt);
```
Ghoulboy78, line 24: Is it necessary to declare the variable?

programs/survival/silk_budding_amethyst.sc
```diff
@@ -0,0 +1,26 @@
...
+    block=='budding_amethyst' &&
+    tool ~ ['diamond_pickaxe', 'netherite_pickaxe'] &&
+    nbt ~ 'silk_touch';
+);
+
+
+__on_player_breaks_block (player, block) -> (
+    __silk_budding_amethyst(player, block) &&
+    (
+        item_nbt = nbt('{Item:{id:"minecraft:budding_amethyst",Count:1b,PickupDelay:10}}');
```
Ghoulboy78, line 23: I recommend adding a little velocity, to make it more like what happens when you actually break a block.

programs/survival/silk_budding_amethyst.sc
DragonCrafted87: nope moved into the spawn call

programs/survival/silk_budding_amethyst.sc
DragonCrafted87: this was fun to figure out if items didn't despawn naturally I have a feeling I'd be randomly finding the ones that flew off like rockets for a while

Ghoulboy78 (approved):
Ok, I think this is quite good rn! Just need to wait for someone else to review.

Firigion (approved):
Looks fine by me!

PR #224, opened 2021-06-25
Create renewable_elytra_trade.sc

Firigion:


programs/survival/renewable_elytra_trade.sc
```diff
@@ -0,0 +1,45 @@
...
+			run(str('/data modify entity %s Offers.Recipes append value %s', e~'command_name', global_trade));
+		);
+		
+		if(	r && //comment this line if adding a trade should always consume an elytra, regardless of if it succeeds or not
+			p~'gamemode'!='creative' && he,
+				inventory_set(p, p~'selected_slot', 0); //consume elytra
+				sound('item.armor.equip_elytra', e~'pos')
+		);
+
+		modify(e, 'tag', 'HasTrade');
```
BisUmTo, line 39: `HasTrade` is a really generic tag
I suggest `renewable_elytra_trade.sc-HasTrade` or something similar

programs/survival/renewable_elytra_trade.sc
```diff
@@ -0,0 +1,45 @@
...
+global_chances = 100;
+
+// Customize the conditions
+_add_trade(e, p) -> (
+	nbt = e~'nbt';
+	if(
+		//e~'dimension'=='the_end' && //villager needs to be in the end
+		nbt:'VillagerData':'level'>0 && // minimum level at which the trade appears
+		he = (p~'holds':0 == 'elytra') && //does the player need to hold an elytra?
+		parse_nbt(nbt:'Tags')~'HasTrade' == null,
```
BisUmTo, line 27: `HasTrade` is a really generic tag
I suggest `renewable_elytra_trade.sc-HasTrade` or something similar

programs/survival/renewable_elytra_trade.sc
```diff
@@ -0,0 +1,45 @@
...
+);
+
+//set the changes of the trade appearing in %
+global_chances = 100;
+
+// Customize the conditions
+_add_trade(e, p) -> (
+	nbt = e~'nbt';
+	if(
+		//e~'dimension'=='the_end' && //villager needs to be in the end
```
BisUmTo, line 24: Adding a config to enable this requirement(?)

Ghoulboy78:
I mean, wouldn't this work better as like a datapack or smth? Cos datapacks are kinda built for this sort of manipulation, so imo they woulld be a better choice to use here.

BisUmTo:
With datapacks you cannot change vanilla trades, but there are a ton of other ways to implement renewable elytras...

programs/survival/renewable_elytra_trade.sc
Firigion: True, will change.

programs/survival/renewable_elytra_trade.sc
Firigion: All conditions should be configurable outside the main code, you mean? I just left "in the end required" off by default because I thought it'd be a good example of how to turn off conditions. I can move them all to a config above.

Firigion:
Yeah, it's just that MelonTech wanted elytra trades, so i implemented it, and I thought it'd be a good option to have among other renewable elytra options.

Datapacks can't edit villager trade pools, they can only summon villagers with set trades. I hope villager trades become data driven like piglin bartering currently is.

programs/survival/renewable_elytra_trade.sc
Ghoulboy78: Well maybe a more descriptive name would be better, like `HasRenewableElytraTrade` to both make it specific to this app and to explain what it means.

programs/survival/renewable_elytra_trade.sc
```diff
@@ -0,0 +1,45 @@
...
+				sound('item.armor.equip_elytra', e~'pos')
+		);
+
+		modify(e, 'tag', 'HasTrade');
+	);
+);
+
+__on_player_interacts_with_entity(player, entity, hand) -> (
+	if(entity=='Leatherworker', _add_trade(entity, player))
+);
```
Ghoulboy78, line 45: Why not use `entity_load_handler`? That way it doesn't get called unnecessarily, but only once on each villager.

Ghoulboy78:
I jsut resolved conflicts

programs/survival/renewable_elytra_trade.sc
Firigion: Because the trade should appear when you first open the villager. Either because you need to right click it with an elytra, because you want the trade to appear at a certain level or only once you move the villager to the end.

I think I'll document better what things this app can do.

programs/survival/renewable_elytra_trade.sc
BisUmTo: I allways prefer to have a way to change config from the chat, wirhout editing the code.

programs/survival/renewable_elytra_trade.sc
Ghoulboy78: I also suggest putting this check at the front, so it fails immediately for villagers which already have it, thus not having to  do redundant checks/calculations

programs/survival/renewable_elytra_trade.sc
Ghoulboy78: Ah I c, ye best to document it better imo.

programs/survival/renewable_elytra_trade.sc
Firigion: Will add this a, long with better documentation on how to customize the app.

Firigion:
> I allways prefer to have a way to change config from the chat, wirhout editing the code.

I don't like the user to be able to edit core mechancis of an app with a command, only stuff taht's meant to be mutable... Idk, if other people agree with bisumto, then I'll add the commands.

programs/survival/renewable_elytra_trade.sc
```diff
@@ -0,0 +1,65 @@
...
+//////////
+
+// Customize the conditions
+_add_trade(e, p, ) -> (
+	nbt = e~'nbt';
+	if(
+		parse_nbt(nbt:'Tags')~'HasRenewableElytraTrade' == null && //already has trade
+		(!global_in_the_end || e~'dimension'=='the_end') && //villager needs to be in the end
+		nbt:'VillagerData':'level'>=global_min_level && // minimum level at which the trade appears
+		(!global_hold_elytra || p~'holds':0 == 'elytra' ) //does the player need to hold an elytra?
```
Ghoulboy78, line 42: Well perhaps these checks should be in a config or smth above, not below the `Stop looking, none of your business here` comment?

programs/survival/renewable_elytra_trade.sc
```diff
@@ -0,0 +1,65 @@
...
+		,
+
+		if(r = (rand(100)<global_chances),
+			run(str('/data modify entity %s Offers.Recipes append value %s', e~'command_name', global_trade));
```
Ghoulboy78, line 46: isn't there a native scarpet method of doing this with `modify(e, 'nbt_modify')` or something? That would be faster cos doesn't have to go through the whole command thing, just directly changes its nbt.

Firigion: Not if you want to make sure that the nbt appends to the end of the list. Maybe there's a way to make sure, but I didn't dig deep. It's fast enough anyway, it's not like it's gonna run many times per tick or anything.

Ghoulboy78: k

programs/survival/renewable_elytra_trade.sc
Ghoulboy78: Oh wait, I'm dumb, sorry nvm.

programs/survival/renewable_elytra_trade.sc
```diff
@@ -0,0 +1,65 @@
...
+global_trade = encode_nbt(
+	{
+		'sell' -> {'id' -> 'minecraft:elytra', 'Count'->1},
+		'buy' -> {'id' -> 'minecraft:emerald', 'Count' -> 40},
+		'buyB' -> {'id' -> 'minecraft:phantom_membrane', 'Count' -> 5},
+		'uses' -> 0,
+		'demand' -> -20,
+		'priceMutiplier' -> 0.0,
+		'specialPrice' -> 0,
+		'rewardExp' -> 1, // boolean (gives/doesn't give xp)
```
Ghoulboy78, line 26: Maybe you could add this as config option, so like have a `global_should_give_xp = true` or smth, and change this line to `'rewardExp' -> number(global_should_give_xp )`?

programs/survival/renewable_elytra_trade.sc
```diff
@@ -0,0 +1,65 @@
...
+				inventory_set(p, p~'selected_slot', 0); //consume elytra
+				sound('item.armor.equip_elytra', e~'pos')
+		);
+
+		modify(e, 'tag', 'HasRenewableElytraTrade'); // give tag even if it failed by random chance
+	);
+);
+
+__on_player_interacts_with_entity(player, entity, hand) -> (
+	if(entity=='Leatherworker', _add_trade(entity, player));
```
Ghoulboy78, line 64: I'm not sure that this will work 100% of the time on servers set to a language other than en_us, I think (not sure, would need to test). In that case, you would need to use `entity~'type'=='villager'` and then another check (I can't remember its name) to see the villager type.

programs/survival/renewable_elytra_trade.sc
Firigion: Hm, I'm not sure I like it. the configs decide how the appearing of the trade behaves. The trade editing is separate and should only be messed with if you have a minimal understanding of vanilla commands/how trades work. I don't want to mix the trade characteristics with the conditions for it to appear.

Firigion: Except for price, all values there are default for a custom trade, but could be modified for a more interesting experience, if you know what you are doing.

programs/survival/renewable_elytra_trade.sc
Firigion: Ohh, I'll look it up and change that. Thanks!

Firigion: Items names do get auto translated, right? I mean, the name we refere to them by in the code is just the code name, the game translates them afterwards?

Firigion: Fixed in f93bf5e.

programs/survival/renewable_elytra_trade.sc
BisUmTo: I think it's a nice decision split appearing condition and the trade.

BisUmTo (approved):

altrisi (approved):
Looks good, works as expected!

I'd make it have (a high) fail chance by default (looks really op to me else, even with that), but since it's customizable it doesn't matter that much.

PR #228, opened 2021-06-26
Update resource_app.sc

altrisi:
This is a complement to gnembon/fabric-carpet#956, updating the example app here according to the changes made in base carpet.


altrisi:
Would appreciate approving reviews, since the old one is no longer valid since there were breaking changes in the Carpet side.

Firigion (approved):
It's ready to merge, right?

altrisi:
Once it gets another approval, yes, since the Carpet PR has already been released.

Ghoulboy78 (approved):

PR #229, opened 2021-06-27
fix lava_sponge.sc

B-2U:
fixed critical bug of the script in the last commit before adding pull request merged,
forgot to rewrite the conditions

### DIFF
--- a/programs/survival/lava_sponge.sc
+++ b/programs/survival/lava_sponge.sc
@@ -10,7 +10,7 @@ __config() -> (
 
 
 __on_player_places_block(player, item_tuple, hand, block)->(
-    if(hand == 'mainhand' || item_tuple:0 == 'sponge' || all(neighbours(pos(block)),_=='lava'),
+    if(hand == 'mainhand' && item_tuple:0 == 'sponge' && neighbours(pos(block))~'lava' != null,
         set(pos(block),'wet_sponge');
         count = 0;
         recursion = 0;


altrisi (approved):
I'm approving it without launching it because I've seen how critical it is and have read reports from users saying it works now.

I don't like doing that, but I guess it's the better way.

Next time please test before approving! This is not the first tiny thing that gets a green review that shouldn't, though at least this is the only one that got merged.

programs/survival/lava_sponge.sc
```diff
@@ -10,7 +10,7 @@ __config() -> (
 
 
 __on_player_places_block(player, item_tuple, hand, block)->(
-    if(hand == 'mainhand' || item_tuple:0 == 'sponge' || all(neighbours(pos(block)),_=='lava'),
+    if(hand == 'mainhand' && item_tuple:0 == 'sponge' && neighbours(pos(block))~'lava' != null,
```
Ghoulboy78, line 13: You could use the suggestion I made here: https://github.com/gnembon/scarpet/pull/220#discussion_r657917284
It appears that you half-implemented it, i.e you changed the whole thing to be nested in an if-loop, but then didn't copy the exact boolean check condition I wrote: `hand=='mainhand' && item_tuple:0=='sponge' && any(neighbours(pos(block)), _=='lava')`

Ghoulboy78: Agh nvm Ill do it mself

Ghoulboy78 (approved):
Ok, the bug is properly gone, so this can be merged

BisUmTo:
`any()` function doesn't exist...

PR #232, opened 2021-07-06
Add Scarpet Edit Installer

altrisi:
Adds an installer for [Scarpet Edit](https://github.com/Ghoulboy78/Scarpet-edit).

The installer works by declaring SE as a dependency and checking before and after whether the app was actually installed.

Depends on Ghoulboy78/Scarpet-edit#85 getting into a release with the language files in it.

Testing is difficult. Sorry. But it Works on my machine™.

### DIFF
--- a/programs/building/scarpet_edit_installer.sc
+++ b/programs/building/scarpet_edit_installer.sc
@@ -1,0 +1,70 @@
+// Scarpet app to download the Scarpet Edit (https://github.com/Ghoulboy78/Scarpet-edit) and its translations.
+// This app is not intended to be downloaded manually, and will only work if downloaded using the "/script download building/scarpet_edit_installer.sc" command
+// to download it from the app store.
+
+printf(text) -> (
+    print(format(text));
+);
+
+checkText = 'Check the Scarpet Edit license at https://github.com/Ghoulboy78/Scarpet-edit/blob/master/licence.txt';
+global_langs = ['es_la', 'it_it'];
+
+if (length(list_files('', 'text')) == 0,
+    firstTime = true;
+);
+
+genResources() -> (
+    resources = [];
+    for (global_langs,
+        resources += {
+            //'source' -> 'https://github.com/Ghoulboy78/Scarpet-edit/releases/latest/download/' + _ + '.json' TODO Change on SE release
+            'source' -> 'https://raw.githubusercontent.com/Ghoulboy78/Scarpet-edit/master/se.data/langs/' + _ + '.json'
+        };
+    );
+    resources += {
+        'source' -> 'https://raw.githubusercontent.com/Ghoulboy78/Scarpet-edit/master/licence.txt'
+    };
+    resources
+);
+
+__config() ->
+{
+    'scope' -> 'global',
+    'stay_loaded' -> false, // Don't keep it running if autoloaded
+    'requires' -> {
+        'carpet' -> '>=1.4.42'
+    },
+    'libraries' -> [
+        {
+            'source' -> 'https://github.com/Ghoulboy78/Scarpet-edit/releases/latest/download/se.sc'
+        }
+    ],
+    'resources' -> genResources()
+};
+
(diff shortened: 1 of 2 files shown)


programs/building/scarpet_edit_installer.sc
```diff
@@ -0,0 +1,70 @@
...
+    'libraries' -> [
+        {
+            'source' -> 'https://github.com/Ghoulboy78/Scarpet-edit/releases/latest/download/se.sc'
+        }
+    ],
+    'resources' -> genResources()
+};
+
+if (length(list_files('', 'text')) != 0 && firstTime,
+    printf('l Scarpet Edit installed successfully (unless otherwise noted)');
```
Ghoulboy78, line 46: What do you mean here by `unless otherwise noted`?

Firigion:
How do I test this? Can I make `/script download` point to a specific branch in the repo?

altrisi:
Nope, you can't (limitations on how Carpet manages it). You'd need to fork the repo, pull this to the master branch, change the `scriptsAppStore` (or similar) rule to point to your repo (see how the default one is done for syntax) and then `/script download` it. However, language installation won't work since it needs an in-dev version of Scarpet Edit (you could change the library to point to the SE PR branch instead of the release if you wanted to test that part too).

Still a draft though, basically because of the language thing not being released yet.

PS: You can also technically modify Carpet to add a StoreSource to every app that doesn't have one (that's how I tested it, can break outside of testing) and then loading it normally.

Firigion:
Okay, I'll wait, then.

programs/building/scarpet_edit_installer.sc
Ghoulboy78: Ah ok

Ghoulboy78:
Yo @altrisi, here's the deal. I could make a new Scarpet-Edit release for this new language installer thing, but the thing is that literally 3 commits have passed since the last release, and it feels kinda awkward to be making a new release. However, if it's absolutely necessary for this, then I can do it (in like a 1.3.1 release or something).
Could you please let me know then whether or not a release containing the changes in Ghoulboy78/Scarpet-Edit#85 is necessary for this to go through?

altrisi:
A new release is not necessary, only adding the language files to the current one to prevent it failing (just languages won't be installed until then).

I'm also planning to make one or two PRs there in the near future, one being automated language selection.

Also this is on hold basically while waiting for release or a pair more changes there, no rush.

Firigion:
~~Just fix https://github.com/Ghoulboy78/Scarpet-edit/issues/62 and we can do 1.4 release.~~

Ghoulboy78:
> Only adding the language files to the current one to prevent it failing (just languages won't be installed until then).

Do you mean adding the language files to the release page along with the app itself? If so I can easily do that.

programs/building/scarpet_edit_installer.sc
```diff
@@ -0,0 +1,70 @@
+// Scarpet app to download the Scarpet Edit (https://github.com/Ghoulboy78/Scarpet-edit) and its translations.
+// This app is not intended to be downloaded manually, and will only work if downloaded using the "/script download building/scarpet_edit_installer.sc" command
+// to download it from the app store.
+
+printf(text) -> (
+    print(format(text));
+);
+
+checkText = 'Check the Scarpet Edit license at https://github.com/Ghoulboy78/Scarpet-edit/blob/master/licence.txt';
+global_langs = ['es_la', 'it_it'];
```
Ghoulboy78, line 10: IMO there should be a better way of accessing the languages which are available, maybe by seeing all the files in the release itself using GithubAPI? Tho that would require adding HTML querying to scarpet API (which isn't entirely out of question tbh...)

Ghoulboy78: And to prove the point, I made this: https://github.com/gnembon/fabric-carpet/pull/990

Ghoulboy78: Well never mind that...

PR #234, opened 2021-07-11
add sscoreboard app

Dorapower:
functions are described in the comments

### DIFF
--- a/programs/utilities/sscoreboard.sc
+++ b/programs/utilities/sscoreboard.sc
@@ -1,0 +1,86 @@
+//statistic scoreboard interface by Lokdora
+//dynamically create scoreboard based on statistics and assigned it with correct value (when player is online)
+//options to distinguish bots from other players
+//due to the scoreboard value cannot be assigned when player is offline, it is better to use this script to create scoreboard objectives for mid/long-term use (or singleplayer world).
+
+
+__config() -> {
+    'stay_loaded' -> 'true',
+	'scope' -> 'global',
+	'commands' -> {
+		'' -> 'help',
+		'clear' -> 'clear',
+		'<stat> <displayName>' -> ['create','sidebar',false],
+		'<stat> <displayName> <displaySlot>' -> ['create',false],
+		'<stat> <displayName> <displaySlot> <includeBots>' -> 'create'
+	},
+	'arguments' -> {
+		'stat' -> {'type' -> 'term'},
+		'displayName' ->{'type' -> 'string'},
+		'displaySlot' ->{'type' -> 'scoreboardslot'},
+		'includeBots' -> {'type' -> 'bool'}
+	}
+};
+
+entity_load_handler('player',_assign(p,new) -> (
+	if(!global_filter_bot||p~'player_type'!=fake,
+		scoreboard('_ss_core',p~'name',statistic(p~'name',global_category,global_entry)) //player needs to login to update scoreboard values
+	)
+););
+
+help() -> (
+	print('/sscoreboard <stat> <[<bot included>]');
+	print('/sscoreboard clear');
+	print('Supported stat format: \'<global_category>.<global_entry>\' or \'<global_entry>\', in which case the global_category is cosidered as custom.');
+	print('global_category included:');
+	print(' - mined(m);');
+	print(' - crafted(c);');
+	print(' - used(u);');
+	print(' - broken(b);');
+	print(' - picked_up(p);');
+	print(' - dropped(d);');
+	print(' - killed(k);');
+	print(' - killed_by(kb);');
+	print(' - custom(cu).');
+);
+
+create(stat,name,slot,includeBots) ->(
+	scoreboard_remove('_ss_core');
+	global_filter_bot=!includeBots;
+	if(	stat~'\\.', [global_category,tmp,global_entry] = stat~'(.*)(\\.)(.*)',
+		[global_category,global_entry] = ['custom',stat]
+	);
+	if(	
+		global_category=='m'||global_category=='mined',	[global_category,criterion]=['mined','minecraft.mined:'+global_entry],
+		global_category=='c'||global_category=='crafted',	[global_category,criterion]=['crafted','minecraft.crafted:'+global_entry],
+		global_category=='u'||global_category=='used',	[global_category,criterion]=['used','minecraft.used:'+global_entry],
+		global_category=='b'||global_category=='broken',	[global_category,criterion]=['broken','minecraft.broken:'+global_entry],
+		global_category=='p'||global_category=='picked_up',	[global_category,criterion]=['picked_up','minecraft.picked_up:'+global_entry],
+		global_category=='d'||global_category=='dropped',	[global_category,criterion]=['dropped','minecraft.dropped:'+global_entry],
+		global_category=='k'||global_category=='killed',	[global_category,criterion]=['killed','minecraft.killed:'+global_entry],
+		global_category=='kb'||global_category=='killed_by',	[global_category,criterion]=['killed_by','minecraft.killed_by:'+global_entry],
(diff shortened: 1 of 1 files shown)


Dorapower:
should I add the script to the survival category? Since it seems nobody have written the utility readme file

programs/utilities/sscoreboard.sc
```diff
@@ -0,0 +1,86 @@
...
+	},
+	'arguments' -> {
+		'stat' -> {'type' -> 'term'},
+		'displayName' ->{'type' -> 'string'},
+		'displaySlot' ->{'type' -> 'scoreboardslot'},
+		'includeBots' -> {'type' -> 'bool'}
+	}
+};
+
+entity_load_handler('player',_assign(p,new) -> (
```
Ghoulboy78, line 25: Does that even work? Also use `__on_player_connects` event, which is more appropriate here.

programs/utilities/sscoreboard.sc
```diff
@@ -0,0 +1,86 @@
...
+	if(!global_filter_bot||p~'player_type'!=fake,
+		scoreboard('_ss_core',p~'name',statistic(p~'name',global_category,global_entry)) //player needs to login to update scoreboard values
+	)
+););
+
+help() -> (
+	print('/sscoreboard <stat> <[<bot included>]');
```
Ghoulboy78, line 32: Accidental < infront of the [

programs/utilities/sscoreboard.sc
```diff
@@ -0,0 +1,86 @@
...
+	print('/sscoreboard clear');
+	print('Supported stat format: \'<global_category>.<global_entry>\' or \'<global_entry>\', in which case the global_category is cosidered as custom.');
```
Ghoulboy78, line 34: I think here you mean that it's considered custom, not cosidered custom

programs/utilities/sscoreboard.sc
```diff
@@ -0,0 +1,86 @@
...
+	print('global_category included:');
```
Ghoulboy78, line 35: I think here would be it includes

programs/utilities/sscoreboard.sc
```diff
@@ -0,0 +1,86 @@
...
+	scoreboard_remove('_ss_core');
+	global_filter_bot=!includeBots;
+	if(	stat~'\\.', [global_category,tmp,global_entry] = stat~'(.*)(\\.)(.*)',
+		[global_category,global_entry] = ['custom',stat]
+	);
+	if(	
+		global_category=='m'||global_category=='mined',	[global_category,criterion]=['mined','minecraft.mined:'+global_entry],
+		global_category=='c'||global_category=='crafted',	[global_category,criterion]=['crafted','minecraft.crafted:'+global_entry],
+		global_category=='u'||global_category=='used',	[global_category,criterion]=['used','minecraft.used:'+global_entry],
+		global_category=='b'||global_category=='broken',	[global_category,criterion]=['broken','minecraft.broken:'+global_entry],
```
Ghoulboy78, line 57: Idk if it's just me , but I think aligning these would make it easier to read

programs/utilities/sscoreboard.sc
```diff
@@ -0,0 +1,86 @@
...
+		global_category=='p'||global_category=='picked_up',	[global_category,criterion]=['picked_up','minecraft.picked_up:'+global_entry],
+		global_category=='d'||global_category=='dropped',	[global_category,criterion]=['dropped','minecraft.dropped:'+global_entry],
+		global_category=='k'||global_category=='killed',	[global_category,criterion]=['killed','minecraft.killed:'+global_entry],
+		global_category=='kb'||global_category=='killed_by',	[global_category,criterion]=['killed_by','minecraft.killed_by:'+global_entry],
+		global_category=='cu'||global_category=='custom',	[global_category,criterion]=['custom','minecraft.custom:'+global_entry],
+		(print(format('rbu Error:','r Unknown global_category. Please check again.')),return())
+	);
```
Ghoulboy78, line 64: I think instead of having a giant if statement and each time calling `[[global_category,criterion]= bla bla bla`, you could instead do:
```
[global_category,criterion]= if(
    global_category=='m'||global_category=='mined',['mined','minecraft.mined:'+global_entry],
    //all the other conditions
    exit(print(format('rbu Error:','r Unknown global_category. Please check again.')))
)
```
The exit call is essentially the same as the return, but I think faster as it won't bother with returning a value, just halt execution right then and there.

programs/utilities/sscoreboard.sc
```diff
@@ -0,0 +1,86 @@
...
+clear() -> (
+	scoreboard_remove('_ss_core');
+	global_filter_bot = false;
+);
+
+__on_tick() -> (
+	if(global_filter_bot,
+		for(player('*'),if (_~'player_type'=='fake',scoreboard('_ss_core',_~'name',null)))
+	);
+)
```
Ghoulboy78, line 86: player('*') only grabs players in the caller dimension. To grab all players across all dimentions use `player('all')`

Ghoulboy78 (commented):
Quite good, just got some general feedback.

Ghoulboy78:
> should I add the script to the survival category? Since it seems nobody have written the utility readme file

No, this is quite good to remain in the utilities folder, and Imma make an issue about the no readme thing, as you are right about it.

Dorapower:
Fixed mistakes and tested. Thanks for your detailed corrections and suggestions.

programs/utilities/sscoreboard.sc
Dorapower: The event handler works well, I have tested it for a lot of times. And I have tried __on_player_connects at the beginning, but it is not triggered when bot is summoned. I don't know whether it's a bug or fake players don't actually connect to the server.

Dorapower: Oh I ran more tests and found the __on_player_connects actually works, it's just the output of print function isn't showed to me. I have switched to this function instead.

Dorapower (commented):
actually there's dimension problem before
anyway it should be all fine now

programs/utilities/sscoreboard.sc
```diff
@@ -0,0 +1,90 @@
...
+	'scope' -> 'global',
+	'commands' -> {
+		'' -> 'help',
+		'clear' -> 'clear',
+		'<stat> <displayName>' -> ['create','sidebar',false],
+		'<stat> <displayName> <displaySlot>' -> ['create',false],
+		'<stat> <displayName> <displaySlot> <includeBots>' -> 'create'
+	},
+	'arguments' -> {
+		'stat' -> {'type' -> 'term'},
```
Ghoulboy78, line 18: do you think there's any way to possibly figure these out so you can do command suggestions? If you look [here](https://github.com/gnembon/fabric-carpet/blob/master/docs/scarpet/Full.md#command-argument-types) and scroll down, you may be able to use the 'objective' or 'criterion' argument types.

Dorapower: oh, i see now

Dorapower: I don't know what is wrong, but after i changing the term to criterion the function just stopped working. I have done some tests and i found that the create function itself still works when i use the "script in" command to run it, but when i use the generated command it said *An unexpected error occurred trying to execute that command*, and no suggestions pop out  when i am tying the commands in.
If I didn't do something wrong by accident, I believe the only way to do the proper command suggestion is to manually type everything in the script which is too tedious. Although I think the behavior of ‘criterion' type might be a bug, which I am not sure, I decided not to make a change to the script now.

BisUmTo:
In my `Carpet RedCraft Addons` I added the `/stats` command and a `offline_stats()` function to access the offline players stats.
I'll look into this source ;-)

BisUmTo (approved):

BisUmTo:
I think that the script info is missing, but the script now should work fine

Ghoulboy78 (approved):

PR #235, opened 2021-07-12
Add portalorient script

rv3r:
This is a simple utility I wrote to prevent headaches when travelling through Nether portals. Makes the player always face a cardinal direction that is "out" of the portal so you don't have to spin around if you forget which way you're supposed to face for each specific portal in your world. This is accomplished by checking each side of the portal for blocks and making the player face the corresponding side(more air blocks or fewer solid blocks) depending on their setting. In the event of a tie, no side is preferred and the player will simply face their original direction.

### DIFF
--- a/programs/survival/portalorient.sc
+++ b/programs/survival/portalorient.sc
@@ -1,30 +1,31 @@
 __config() ->
 (
-	m(
-		l('scope','player'),
-		l('stay_loaded',true),
+	{
+		['scope','player'],
+		['stay_loaded',true],
 		
-		l('commands',
-			m(
-				l('<mode>','__change')
-			)
-		),
-		l('arguments',
-			m(
-				l('mode',
-					m(
-						l('type','term'),
-						l('options',
-							l('off','air','solid')
-						),
-						l('suggest',
-							l('off','air','solid')
-						)
-					)
-				)
-			)
-		)
-	)
+		['commands',
+			{
+				['<mode>','__change']
+			}
+		],
+
+		['arguments',
+			{
+				['mode',
+					{
+						['type','term'],
+						['options',
+							['off','air','solid']
+						],
+						['suggest',
+							['off','air','solid']
+						]
+					}
+				]
+			}
+		]
+	}
 );
 
 //setup the script's data if nothing is there already
@@ -64,14 +65,14 @@ __on_player_changes_dimension(player, from_pos, from_dimension, to_pos, to_dimen
 		);
 
 		//get the corners of the portal and construct an offset to scan each side
-		offset = l(0,0,0);
+		offset = [0,0,0];
 		offset:(2 - axis) = 1;
 		corners =  __corners(center,offset);
 
 		//scan each side of the portal for blocks based on the player's chosen mode
 		//	air -> face the side with more air blocks
 		//	solid -> face the side with less solid blocks
-		sidelist = l();
+		sidelist = l[];
 		loop(2,
 			sign = 2 * _ - 1;
 			sidelist += volume(corners:0 + sign * offset,corners:1 + sign * offset,
@@ -108,9 +109,9 @@ __on_player_changes_dimension(player, from_pos, from_dimension, to_pos, to_dimen
 //according to profile_expr(), it's about 2.5x faster
 __corners(center,offset) ->
 (
-	corners = l();
+	corners = [];
 	//use the vector normal to the portal plane to construct a vector along the portal plane 
-	corneroffset = l(1,1,1) - offset;
+	corneroffset = [1,1,1] - offset;
 	//we'll need to check along each direction of the diagonal to find both corners
 	loop(2,
 		//-1 for negative direction, +1 for positive direction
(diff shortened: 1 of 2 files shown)


altrisi:
Haven't read the code yet, just the summary, but isn't that done by vanilla? IIRC portals count the amount of non-air blocks on each side and decide how to face the player depending on that and where you were facing when you entered.

altrisi:
Oh, just read the "always" word, so this one makes it so even if you are facing the opposite way when you enter you'd appear looking into the open face anyway?

rv3r:
> Haven't read the code yet, just the summary, but isn't that done by vanilla? IIRC portals count the amount of non-air blocks on each side and decide how to face the player depending on that and where you were facing when you entered.

Perhaps this is the case, but I'm can't seem to replicate this behavior. Edited summary to note that player will always face a cardinal direction.

rv3r:
After a few attempts, I realize now what you meant by the vanilla behavior, but I can't get it to trigger consistently.

I've attached some gifs showing the vanilla behavior I'm seeing when going through two portals oriented along different axes, but keep it mind that it doesn't seem to be any better with aligned portals.
I also have gifs showing script behavior to show the purpose of the script.

Facing left side of portal

vanilla
portalleftgif

script
scriptleftgif

Facing right side of portal

vanilla
portalrightgif

script
scriptrightgif

Facing into portal

vanilla
portalstraightgif

script
scriptstraightgif

altrisi:
Well yes vanilla is always bugged so yes. The app makes sense anyway, will review it when I have a second and remember about it.

programs/survival/portalorient.sc
```diff
@@ -0,0 +1,186 @@
...
+	)
+);
+
+//setup the script's data if nothing is there already
+//it's an empty map, so players without data will have a null mode
+__on_start() ->
+(
+	if(!load_app_data(),
+		store_app_data(m());
+	)
```
Ghoulboy78, line 36: Just for the sake of curiosity, hw hard would it be to automatically make the players be in off mode instead of saving an empty map? (Im genuinely asking cos I have no clue how tf store_app_data works and whether it will overwrite other player's data or otherwise, idk)

programs/survival/portalorient.sc
```diff
@@ -0,0 +1,186 @@
...
+			//length(check) - check ~ edge to prevent checking too many blocks
+			edges = map(l(range(length(check) - check ~ edge)),block(_ * listdirection * direction + pos)),
+		   	//if we're already at the corner after the first check, we only need one block
+			edges = l(block(pos));
+		);
+
+		//find which blocks along the edges are actually the corners
+		corners:_ = __lastportal(edges,null,null);
+	);
+	return(corners);
```
Ghoulboy78, line 159: Don't need return call here, just leaving `corners` will return it

programs/survival/portalorient.sc
rv3r: It should only write the empty map if there's no app data already stored, so what gets overwritten is just `null`; no player data can be overwritten(in theory, but I can't think of any edge cases).

rv3r: I suppose `false` can also get overwritten, but either way a map with non-null keys and values should pass the check and `!load_app_data()` should return `false`, preventing `store_app_data()` from getting called.

Ghoulboy78: Ah ok

BisUmTo (approved):

BisUmTo:
I'm loving this app! 💛

altrisi:
Note: Since 1.4.35, using `return` _at the end_ of a function is optimized at compile time, so it has the same overhead as without it, so choose the one you like the most. `return` at any other position is still unoptimized though.

PR review Coming Soon™.

programs/survival/portalorient.sc
```diff
@@ -0,0 +1,186 @@
...
+						),
+						l('suggest',
+							l('off','air','solid')
+						)
+					)
+				)
+			)
+		)
+	)
+);
```
Firigion, line 28: Think about changing this for the modern syntax that uses `[ ]` and `{ }` for lsits and maps respectively.

programs/survival/portalorient.sc
```diff
@@ -0,0 +1,186 @@
...
+(
+	m(
+		l('scope','player'),
+		l('stay_loaded',true),
+		
+		l('commands',
+			m(
+				l('<mode>','__change')
+			)
+		),
```
Firigion, line 11: Add a command with no argumetns that prints info on the three modes and maybe alos what the app does exactly? Since apps are now meant to be downloaded directly from in-game, we souldn't rely on people reading the code to understand the apps.

Firigion (changes requested):
Other than these two things, it looks good. The style comment I don't care much about. I'd be glad if you implemented it. The help command comment I would like to see implemented.

programs/survival/portalorient.sc
rv3r: I think I got them all.

programs/survival/portalorient.sc
rv3r: Gave it a first attempt, please comment on the language in case anything is unclear. It's effectively a condensed copy of the docs description.

programs/survival/portalorient.sc
```diff
@@ -0,0 +1,206 @@
...
+		
+		['commands',
+			{
+				['','__guide'],
+				['<mode>','__change']
+			}
+		],
+
+		['arguments',
+			{
```
Firigion, line 15: Proper map syntax would be
```
{
    'scope' -> 'player',
    'stay_loaded' ->true,
    'commands' -> {
        '' -> '__guide',
       '<mode>' -> '__change',
    },
    'arguments' -> { ...

```
and so on

Firigion (approved):
Alright, just a minor syntax thing.

programs/survival/portalorient.sc
rv3r: I've stopped using `m()` and `l()` because they're "older", but the docs call my method here "equivalent" to yours and I prefer to do it this way.

PR #240, opened 2021-08-04
Fix mandelbrot_segment() definition to properly render entire canvas.

aebkea:
Just what the title says. It was a one line change. Though I also removed the superfluous empty first line.

### DIFF
--- a/programs/simulations/sequential_mandelbrot_painter.sc
+++ b/programs/simulations/sequential_mandelbrot_painter.sc
@@ -40,5 +40,5 @@ mandelbrot_segment_offset(x1,y1,z1, size, center_x, center_z, radius, do_log, xs
    )
 );
 
-// /script invokepoint mandelbrot_segment_offset x y z 100 -0.74515346 0.11259498 0.007 false 50 0
+// /script in sequential_mandelbrot_painter invokepoint mandelbrot_segment_offset x y z 100 -0.74515346 0.11259498 0.007 false 50 0
 // Obviously you would normally do that for much larger area, like 0 1 0 10000 ....
--- a/programs/simulations/sequential_mandelbrot_painter.sc
+++ b/programs/simulations/sequential_mandelbrot_painter.sc
@@ -1,8 +1,7 @@
-
 mandelbrot_whole(x1,y1,z1,size) -> mandelbrot_segment(x1,y1,z1,size, 0, 0, 2, true);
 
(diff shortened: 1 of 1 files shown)


Ghoulboy78:
Can yo please explain what exactly was wrong with the previous behaviour, and how this fixes it? (preferably with screenshots)

aebkea:
@Ghoulboy78 `size` refers to the size of the complete Mandelbrot set image Minecraft's block coordinate system. `radius` refers to the size of the Mandelbrot set in internal coordinate system used by the `Mandelbrot` function. As an example for why we have the distinction, when all works well, fixing `size` and increasing `radius` will zoom in the image without changing the size it takes up in the world.

As it was before, `radius` was used instead of `size` as part of an argument `xsteps` which determines how many columns of image are generated. The intended behavior is for `mandelbrot_segment()` to render a potentially offset and zoomed Mandelbrot set on a square canvas, but with `xsteps=2*radius+1` we would, for common radii, step through too few columns. We need to step through `2*size+1` columns to match the height of the columns and make it a square.

In short, the previous behavior would be that `mandelbrot_whole()` and `mandelbrot_segment()` would generally render only 5 columns (if using the common radius of 2), regardless of the value of `size`, whereas now it renders the whole square (with side length `2*size+1`).

aebkea:
I've just noticed that I actually over-set the `xstep` by 1. This was caught and handled gracefully by the program, but it's fixed now anyways. I also changed the indexing of the column log which prints out to tell you the progress to start at 1 instead of 0. This makes more sense in the [column]/[side length] format the log is given in.

Ghoulboy78 (approved):
Thanks for the explanation mate.

BisUmTo (approved):
Running the command at the end of the script file, I get "Error while evaluating expression: UNDEFINED".
Using the command `/script in sequential_mandelbrot_painter run mandelbrot_segment_offset(0,1,0,100,-0.74515346,0.11259498,0.007,false,200,0)` it works fine for me.

BisUmTo:
@Ghoulboy78 do you know why does this error pop up?
> "Error while evaluating expression: UNDEFINED"

altrisi:
`UNDEFINED` is thrown when a legacy invocation fails to find a function name. I think this is because this was originally thought to be added to the chat scope (via `/script run [all the file]`, when apps wouldn't come from files), and then it would be in the chat's scope. It should work by changing it to `/script in sequential_mandelbrot_painter invokepoint ...`

PR #243, opened 2021-08-29
fixed callback error from lava_sponge.sc

Crec0:


altrisi (approved):
Tested it, works correctly.

PLEASE don't push changes after approvals then instamerge, it removes the whole point of the review system. In this case this goes for you @Ghoulboy78, as this got introduced in ([`8f75289` (#229)](https://github.com/gnembon/scarpet/pull/229/commits/8f75289f487e6f561c5599fb49a4bec9f8805d3b)), but this should apply to anyone who can merge.

Firigion (approved):
simple enough change

PR #244, opened 2021-09-04
Created UUIDs library

BisUmTo:
Functions:
- generate_uuid()
- generate_uuid_string()
- uuid_to_string(uuid)
- string_to_uuid(string)

Included utility functions:
- hex_to_number(string)
- or_bitwise(a, b)
- and_bitwise(a, b)

### DIFF
--- a/programs/utilities/uuids.scl
+++ b/programs/utilities/uuids.scl
@@ -1,0 +1,82 @@
+// Returns a new UUID int array
+generate_uuid() -> (
+    uuid = map(range(4), floor(rand(2^32)));
+    uuid:1 = _and_bitwise(uuid:1, 0xffff0fff); // clear version
+    uuid:1 = _or_bitwise(uuid:1, 0x00004000);  // set to version 4
+    uuid:2 = _and_bitwise(uuid:2, 0x3fffffff); // clear variant
+    uuid:2 = _or_bitwise(uuid:2, 0x80000000);  // set to IETF variant
+    uuid
+);
+
+// Returns a new UUID string
+generate_uuid_string() -> uuid_to_string(generate_uuid());
+
+// Converts UUID int array to UUID string
+uuid_to_string(uuid) -> (
+    if( type(uuid) != 'list' ||
+        length(uuid) != 4 ||
+        reduce(uuid, _a || type(_) != 'number', false),
+    // then
+        exit('Invalid UUID: ' + uuid)
+    );
+    str('%08x-%04x-%04x-%04x-%04x%08x', uuid:0, floor(uuid:1 / 2^16), uuid:1 % 2^16, floor(uuid:2 / 2^16), uuid:2 % 2^16, uuid:3)
+);
+
+// Converts UUID string to UUID int array
+string_to_uuid(uuid_string) -> (
+    if( type(uuid_string) != 'string' ||
+        c = split('-', uuid_string); 
+        length(c) != 5 ||
+        length(c:0) != 8 ||
+        length(c:1) != 4 ||
+        length(c:2) != 4 ||
+        length(c:3) != 4 ||
+        length(c:4) != 12,
+    // then
+        exit('Invalid UUID string: ' + uuid_string)
+    );
+    c = [c:0, c:1 + c:2, c:3 + slice(c:4, 0, 4), slice(c:4, 4, 12)];
+    try(
+        map(c, _hex_to_number(_)),
+    // catch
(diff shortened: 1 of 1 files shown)


BisUmTo:
Should it go into `fundamentals` ?

Ghoulboy78:
yo bisumto, numbers in scarpet are stored in longs, so you can go with full 64 bits

programs/utilities/uuids.scl
```diff
@@ -0,0 +1,82 @@
...
+    loop(32,
+        c += c;
+        if(a >= 2^31 && b >= 2^31, c += 1);
+        a += a;
+        b += b;
+        a = a % 2^32;
+        b = b % 2^32;
+    );
+    c
+);
```
Ghoulboy78, line 82: IMO this should just be added to scarpet itself. as it would make it much faster that way.

Ghoulboy78: You know what, Imma add it myself.

programs/utilities/uuids.scl
```diff
@@ -0,0 +1,82 @@
...
+    char = split('', lower(string));
+    for(char,
+        c = c * 16; 
+        t = global_hex ~ _;
+        if(t == null, exit('Invalid hexadecimal character \'' + _ + '\' in string: ' + string));
+        c += t
+    );
+    c
+);
+
```
Ghoulboy78, line 58: This already exists in math.sc:
https://github.com/gnembon/scarpet/blob/master/programs/fundamentals/math.sc#L23

Ghoulboy78:
> Should it go into `fundamentals` ?

IMO it should, I think that should be our 'library' folder (i.e the folder for libraries)

programs/utilities/uuids.scl
BisUmTo: `hex(num)` converts num into hex string (like `str('%x',num)`)
this function converts a hex string into a number

Ghoulboy78: I see. Then maybe this function should be added to math.sc, and that renamed to math.scl

Ghoulboy78:
ok, the bitwise operations can be removed once gnembon/fabric-carpet#1094 is merged.

Ghoulboy78:
Adding label cos its technically waiting on #1094

rv3r:
I think the label should be removed now that gnembon/fabric-carpet#1094 has been merged. This just needs to be moved to fundamentals then?

rv3r:
uuid
Looks like hex precision isn't working and is instead reporting the entire number? carpet 1.4.51

BisUmTo:
I think I found the issue: I'm assuming that the number is an integer while making bitwise operations.
It is a double/long instead...

Parietic (approved):
Looks and works great so far.
Would love for this to be a scarpet feature, but this is a nice option until then.

BisUmTo:
Does it works? I'm not sure 😅
It didn't work last time I tested it (when I added the mhh label)

Parietic:
I mean its working pretty well for me in carpet 1.4.56.

image
Version identifier is correct, so is variety.

BisUmTo:
I'm surprised 😅

programs/utilities/uuids.scl
```diff
@@ -0,0 +1,57 @@
...
+    if( type(uuid_string) != 'string' ||
+        c = split('-', uuid_string); 
+        length(c) != 5 ||
+        length(c:0) != 8 ||
+        length(c:1) != 4 ||
+        length(c:2) != 4 ||
+        length(c:3) != 4 ||
+        length(c:4) != 12,
+    // then
+        exit('Invalid UUID string: ' + uuid_string)
```
Ghoulboy78, line 36: If this is going to be a library, then maybe it's best to throw an error so user can catch it?

programs/utilities/uuids.scl
```diff
@@ -0,0 +1,57 @@
...
+// Returns a new UUID string
+generate_uuid_string() -> uuid_to_string(generate_uuid());
+
+// Converts UUID int array to UUID string
+uuid_to_string(uuid) -> (
+    if( type(uuid) != 'list' ||
+        length(uuid) != 4 ||
+        reduce(uuid, _a || type(_) != 'number', false),
+    // then
+        exit('Invalid UUID: ' + uuid)
```
Ghoulboy78, line 20: Same here, this being largely cos that way if it's being used within the program then it won't explode if the programmer sends a wrong input, especially given that it's a ibrary

PR #249, opened 2021-10-01
Fixes #248 by checking for matching potion effect when item was potion

### LINKED ISSUE #248: shulkerboxes.sc script "restock same" can restock with different potions
From SouthernPixel in the #carpet-mod channel, "...when I use my splash potion of fire resistance it brings a potion of swiftness," that is, the script gets the first potion, not the first matching potion. This is likely because the item id("potion") matches but the nbt tag describing the potion effect is not checked.

Expected behavior: when a potion leaves an inventory slot, "restock same" shulker box restocks slot with matching potion type
Actual behavior: first potion found in box is used to restock the slot

Reproduction steps
1. Get shulkerboxes.sc script
2. Put multiple potion types in a shulker box
3. Rename box to "restock same"
4. Have box in inventory and throw or drop a potion
5. Potion type that restocks slot is not necessarily previous potion type

Carpet 1.4.46
MC 1.17.1

rv3r:
In the event that an item was a potion, as long as box mode is "same" or "keep", requires matching potion effect for item to be restocked.

### DIFF
--- a/programs/survival/shulkerboxes.sc
+++ b/programs/survival/shulkerboxes.sc
@@ -291,9 +291,11 @@ __swap_stack(player, slot, previous_item, item, count, tag) ->
             if (type(shulker_stacks)=='nbt', shulker_stacks = [shulker_stacks]);
             sb_item_count = length(shulker_stacks);
             for( shulker_stacks,
-               if( _:'id' == item_fqdn,
+					// item matches and not a potion or matching potion effect as well as same and keep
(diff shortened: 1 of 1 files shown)


Firigion (commented):
Doesn't work for me. Works for placing blocks, not for drinking potions. Do I need to do something specific to make it work? Idk what the previous comment was about. I think I thought it was another PR.

rv3r:
Drop a regular potion or throw a splash potion. It doesn't restock the slot because an item(empty bottle) is left after drinking a potion

Firigion:
Woops, of course. Would be neat tho, eh? :P

Firigion:
Still waiting for one more approval. @Ghoulboy78?

Ghoulboy78 (approved):

PR #251, opened 2021-10-06
sleep in script

nsluhrs:
added a script that allows players to skip to night by right clicking a bed

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 *.sc linguist-language=Python
+*.scl linguist-language=Python


programs/survival/sleep_in.sc
```diff
@@ -0,0 +1,34 @@
...
+  )
+);
+
+_skip_day() -> (
+	//dtime = day_time()%24000; //day time
+	if((dtime = day_time()%24000)< global_night_time, // if it's sleepy time
+		current = day_time();
+		day_time(current + (global_night_time - dtime) );
+		print('day skipped')
+	);
```
Firigion, line 33: This dpoesn't work if it's already night. Again, if that's intended, disregard this, but I'm not sure that's intended.

Firigion:
Ah, also: think about making the message  abit more visible. Could be a message over the hotbar (I forge tthe name), a title screen or a coloured chat message.

programs/survival/sleep_in.sc
Ghoulboy78: well I think it is intended, seen as though this script is trying to skip the day. If it's night, then there is no need to skip the day.

Ghoulboy78 (approved):
Looks pretty good to me

nsluhrs:
Closed this by mistake but as long as I'm remaking it will make change recommended by Firigion then resubmit

PR #254, opened 2021-10-10
Add player breaks event call to hammer and veinminer

Ghoulboy78:
This allows them to interface with `auto_pickup.sc`, which I've found is immensely helpful, seen as though these apps would tend to be used together.

### DIFF
--- a/programs/survival/hammer.sc
+++ b/programs/survival/hammer.sc
@@ -1,10 +1,5 @@
 //!scarpet v1.5
 
-// stay loaded
-__config() -> {
-	'stay_loaded'->true
-};
-
 __get_block_range(position, face) -> 
 (
 	l(x,y,z) = position;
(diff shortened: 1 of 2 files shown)


programs/survival/hammer.sc
```diff
@@ -1,11 +1,9 @@
...
 
 // stay loaded
-__config() -> (
-   m(
-      l('stay_loaded','true')
-   )
-);
+__config() -> {
+	'stay_loaded'->true
+};
```
Firigion, line 7: Isn't this redundant, now that stay loaded is default?

programs/survival/veinminer.sc
```diff
@@ -1,9 +1,7 @@
 // stay loaded
-__config() -> (
-   m(
-      l('stay_loaded','true')
-   )
-);
+__config() -> {
+	'stay_loaded'->true
+};
```
Firigion, line 4: same

BisUmTo (approved):

PR #262, opened 2021-11-13
Updated and improved flower.sc

manyrandomthings:
* Added support for 1.18 (still works for pre-1.18 versions)
* Now uses tasks to not freeze the game while placing blocks
* Added a count command that counts the number of flower in a given area
* Now requires OP to fill blocks
* Added settings:
  * ignore_biome - doesn't check for flower forest biome when filling/counting
  * glass_mode - places glass blocks instead of flowers (mainly for showing 3D areas)
  * place_air_above - place air above flowers (only when glass_mode is false)

### DIFF
--- a/programs/utilities/flower.sc
+++ b/programs/utilities/flower.sc
@@ -1,84 +1,213 @@
-//
-// Places the flower pattern that bonemeal creates when in flower forests.
-// large areas may cause quite a bit of lag to run.
-// Only places grass blocks + flowers in flower forest biomes.
-//
-// command usage:
-// /flower <position> print - prints what flower will generate at given position
-// /flower <position> create <radius> - places grass block with flowers on top in given radius (square area)
-// /flower <position> create <radius> only <flower> - same as create, but only places the chosen flower type
-//
+// Utility for the flower pattern in flower forests
+// May cause quite a bit of lag if run on a large area
 
-// flowers that generate in flower forest biome (in order)
-global_FLOWERS = ['dandelion', 'poppy', 'allium', 'azure_bluet', 'red_tulip', 'orange_tulip', 'white_tulip', 'pink_tulip', 'oxeye_daisy', 'cornflower', 'lily_of_the_valley'];
+// use legacy noise for versions before 1.18
+global_use_legacy_noise = system_info('game_major_target') < 18;
+
+// list of flowers that generate in flower forest biome (in order)
+global_flower_list = [
+  'dandelion',
+  'poppy',
+  'allium',
+  'azure_bluet',
+  'red_tulip',
+  'orange_tulip',
+  'white_tulip',
+  'pink_tulip',
+  'oxeye_daisy',
+  'cornflower',
+  'lily_of_the_valley'
+];
+global_flower_count = length(global_flower_list);
+
+// glass blocks that have a similar color to each of the flowers
+global_glass_list = [
+  'yellow_stained_glass',     // dandelion
+  'brown_stained_glass',      // poppy
+  'magenta_stained_glass',    // allium
+  'light_gray_stained_glass', // azure_bluet
+  'red_stained_glass',        // red_tulip
+  'orange_stained_glass',     // orange_tulip
+  'glass',                    // white_tulip
+  'pink_stained_glass',       // pink_tulip
+  'gray_stained_glass',       // oxeye_daisy
+  'blue_stained_glass',       // cornflower
+  'white_stained_glass'       // lily_of_the_valley
+];
+
+// noise constants
+// "seed" and "scale" values are from vanilla_worldgen/worldgen/configured_feature/flower_flower_forest.json
+// scale must be converted to a float then back to double
+// if using a custom noise, "firstOctave" must be 0 and "amplitudes" must be [1.0], otherwise this script doesn't support it
+global_flower_seed = 2345;
+global_scale_1 = 0.02083333395421505;
+global_scale_2 = global_scale_1 * 1.0181268882175227;
+global_double_perlin_amplitude = 0.8333333333333333;
+
+// these seeds are the first 2 nextLong calls of new Random(global_flower_seed) (for 2345: -1223197305642693068, -8087649459364435462)
+// then xor with "octave_0".hashCode() (1261148513)
+// todo: have these values automatically calculated from global_flower_seed
+global_perlin_seed_1 = -1223197304453310635;
+global_perlin_seed_2 = -8087649458443489125;
+
+// settings
+global_setting_ignore_biome = false;
+global_setting_glass_mode = false;
+global_setting_place_air_above = true;
 
 __config() -> {
   'stay_loaded' -> true,
   'commands' -> {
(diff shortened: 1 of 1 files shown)


programs/utilities/flower.sc
```diff
@@ -1,140 +1,299 @@
...
+// these seeds are the first 2 nextLong calls of new Random(global_flower_seed) (for 2345: -1223197305642693068, -8087649459364435462)
+// then xor with "octave_0".hashCode() (1261148513)
+// todo: have these values automatically calculated from global_flower_seed
+global_perlin_seed_1 = -1223197304453310635;
+global_perlin_seed_2 = -8087649458443489125;
+
+// settings
+global_setting_ignore_biome = false;
+global_setting_glass_mode = false;
+global_setting_place_air_above = true;
```
Crec0, line 56: Since these are constant values, you can name them `global_SCALE_1`, `global_SCALE_2` and so on. 
This will separate them from other global values

Crec0 (commented):
Other than that, it looks good

programs/utilities/flower.sc
manyrandomthings: I checked a bunch of other scripts and none of them used uppercases for constants so I figured I'd just keep these named similarly to those

Ghoulboy78: ye I would agree with manyrandomthings here, best to keep code format the same all across

programs/utilities/flower.sc
```diff
@@ -1,140 +1,299 @@
...
+  dz = abs(pos1 : 2 - pos2 : 2) + 1;
 
-  // place flower
-  set(position, flower);
+  dx * dy * dz; // return volume size
+);
+
+__print_error(message) -> (
+  print(player(), format('r ' + message))
+);
```
Ghoulboy78, line 267: You could just have `exit()` call in here by doing `exit(print(player(), format('r ' + message)))`.

Firigion:
User-wise, it looks brilliant. Love the patterns too
image

Just one suggestion: add a command to print the glass-flower correspondance. You can't expect people to go read the app code once the app is in this repo, since to install it you just need to do `/script download`.  My suggestion would be something like `/flowers glasstypes` to print them all and `/flower glasstypes <flower>` to print a specific flower's glass.

Code wise, I'll take a deeper look tomorrow, but it looks promising.

Also, I'd love for the configs to be interactive, i.e. add a `/flowers settings` command (with no specified setting) and make that print all the current settings. The printed settings would be clickable, like the carpet menu, and would allow you to change the currents value by suggsting comands on click.

This is absolutely up to you, I'm not gonna block this PR if you don't wanna implement it. But if you do, you can ask me for help and/or take a look at my older apps, they implement interctive menus.

Briliant work!

programs/utilities/flower.sc
```diff
@@ -1,140 +1,296 @@
...
+__print_flower_at(position) -> (
+  print(player(), str('Flower at %s: %s', join(', ', position), global_flower_list : _get_flower_id_at(position)));
 );
 
 _get_flower_id_at(position) -> (
-  x = position : 0;
-  z = position : 2;
+  x = floor(position : 0);
+  y = floor(position : 1);
+  z = floor(position : 2);
```
Firigion, line 191: you can do this as `[x, y, z] = position;`. Unpacking is nice.

Firigion (approved):
Looks good to me. I'll mark this review as approval because every comment is regarding the optional addition of interactive settings. There's also cone comment regarding something that may not be needed, but I wouldn't mind accepting the PR even if that were the case. It's up to you if we merge or wait for one more iteration!

Firigion:
Also, don't meadows spawn flowers? Or do they do it like plains, randomized for each locaion, instead of with a deterministic structure?

programs/utilities/flower.sc
```diff
@@ -1,140 +1,296 @@
...
+
+  // check for OP
+  if(p ~ 'permission_level' < 2,
+    __error('You must be OP to fill blocks');
   );
-  print('Filled ' + count + ' flowers');
+  // check if glass_mode is not enabled and height of volume is more than 1
+  if(!global_setting_glass_mode && floor(pos1 : 1) != floor(pos2 : 1),
+    __error('You must have glass_mode setting enabled to fill an area with a height greater than 1');
+  );
```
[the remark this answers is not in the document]
manyrandomthings: An interactive menu sounds pretty nice, but also sounds like a bunch of work so I don't think I'll add that right now. Maybe if I ever come back to this script I'll try to add that

programs/utilities/flower.sc
```diff
@@ -1,140 +1,296 @@
...
+    filter_flower_id = global_flower_list ~ (flower_filter - 'minecraft:');
+
+    // checks if given filter is a flower
+    if(filter_flower_id == null,
+      __error('Not a valid flower: ' + flower_filter);
+    ),
+  //else:
+    filter_flower_id = null;
+  );
+
```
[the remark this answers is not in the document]
manyrandomthings: The problem is if I change `suggest` to `options` there then it starts to suggest every block type, and I'd like it to only suggest the valid flower types

programs/utilities/flower.sc
manyrandomthings: Didn't realize that was in scarpet, that's pretty helpful. Also, these coordinates need to be integers which is why this `floor()` is here, but now that I look at the rest of the code I don't think there's really a way floats could end up here, so I'll assume this is fine to change

manyrandomthings:
Yeah meadows have flowers in a very similar way, but I haven't added that to this for a few reasons:
* Meadows have less flower types than flower forests (only 6 instead of 11), including some parts where it only generates tall grass instead of flowers
* The way it calculates the noise is more complicated (dual_noise_provider instead of noise_provider)
* The blobs of flowers are so much larger, often a whole meadow biome is only one type of flower since the perlin scale is about 48 times larger. So if you see one type of flower, pretty much anywhere around it is still going to be the same type of flower

programs/utilities/flower.sc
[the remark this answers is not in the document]
Ghoulboy78: Well then perhaps you could make it have both `suggest` and `options` parameters so you it will restrict the options.

Firigion (approved):
Alright awsome

Crec0 (approved):

d0x7:
Any update on when, or if, this will be merged? Current version seems broken in 1.18 and probably not everyone searches through existing issues and copies the code manually, but just use ```/download``` as someone mentioned earlier.

altrisi (approved):
Looks great! Sorry for the delay...

Will merge this in a sec, but saw you generate noises on the fly, and was wondering whether the new `sample_noise` function would get those so you don't have to copy the logic and so the app doesn't break if mojang decides to change it again. Also note that modifying the world from off-thread makes every set operation join to the main thread, so the amount of blocks that will be placed will depend on how busy the server is and other stuff (like the server only processing those tasks at the end of the tick, so things that fit in a tick may take multiple because of the moment to execute tasks having finished), I'd consider batching.

Maybe for the next iteration though!

PR #263, opened 2021-11-23
Potion Master Cleric Script

opsaaaaa:
Adds 6 randomized potion trades to the cleric.

### DIFF
--- a/programs/survival/potion_master_cleric.sc
+++ b/programs/survival/potion_master_cleric.sc
@@ -1,5 +1,5 @@
-// This addon adds 6 randomized potion trades to the Cleric. 
-// It makes Haste, Dolphins Grace, Levitation, Wither, Blindness, 
+// This addon adds 7 randomized potion trades to the Cleric. 
+// It makes Haste, Dolphins Grace, Levitation, Wither, Blindness, Luck, 
 // and Mining Fatigue available in survival as potions.
 // Each effect can appear as a regular, splash, or lingering potion.
 // Once a cleric becomes a Journeyman they can developer 0-3 potion trades. 
--- a/programs/survival/potion_master_cleric.sc
+++ b/programs/survival/potion_master_cleric.sc
@@ -1,0 +1,125 @@
+// this addon adds potion trades to journyman clerics
+
+
+// ALl the avalible trades
+global_potion_trades = [
+	'{
+		sell: {id: "minecraft:%s", Count:1, tag:{
+			CustomPotionEffects:[{Duration:9600, Id: 30, Amplifier: 3}], 
+			CustomPotionColor: 5592575, display: {Name:\'{ "text":"Potion of Grace"}\'}}
+		},
+		buy: {id: "minecraft:emerald", Count: %d},
+		buyB: {id: "minecraft:tropical_fish", Count: 1},
+		uses: 0,
+		demand: -20,
+		priceMutiplier: 0.0,
+		specialPrice: 0,
+		rewardExp: 1,
+		xp: 1
+	}','{
+		sell: {id: "minecraft:%s", Count:1, tag:{
+			CustomPotionEffects:[{Duration:9600, Id: 3, Amplifier: 1}], 
+			CustomPotionColor: 16755200, display: {Name:\'{ "text":"Potion of Haste"}\'}}
+		},
+		buy: {id: "minecraft:emerald", Count: %d},
+		buyB: {id: "minecraft:raw_gold", Count: 1},
+		uses: 0,
+		demand: -20,
+		priceMutiplier: 0.0,
+		specialPrice: 0,
+		rewardExp: 1,
+		xp: 1
+	}','{
+		sell: {id: "minecraft:%s", Count:1, tag:{
+			CustomPotionEffects:[{Duration:300, Id: 25, Amplifier: 3}], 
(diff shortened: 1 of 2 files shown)


programs/survival/potion_master_cleric.sc
```diff
@@ -0,0 +1,138 @@
...
+	
+];
+
+global_potion_types = ['splash_potion','lingering_potion','potion'];
+global_potion_emerald_cost = [4,8];
+
+
+_add_trades(cleric, p) -> (
+	// add 0-3 potion trades
+	loop(rand(4),
```
Ghoulboy78, line 106: If you're gonna add randomly between 0 and 3 trades, should also specify that in readme

programs/survival/potion_master_cleric.sc
```diff
@@ -0,0 +1,138 @@
...
+		priceMutiplier: 0.0,
+		specialPrice: 0,
+		rewardExp: 1,
+		xp: 1
+	}'
+	
+];
+
+global_potion_types = ['splash_potion','lingering_potion','potion'];
+global_potion_emerald_cost = [4,8];
```
Ghoulboy78, line 101: If these are customizable, maybe keep them at the very top so they're more visible, cos otherwise they may get lost under the big list right above.

programs/survival/potion_master_cleric.sc
```diff
@@ -0,0 +1,138 @@
...
+		trade = nbt(str(
+			// select a random trade from the list above
+			global_potion_trades:rand(length(global_potion_trades)),
+			// Randomize the potion type 
+			global_potion_types:rand(length(global_potion_types)),
+			// randomize the cost of each trade
+			(global_potion_emerald_cost:0) + rand(global_potion_emerald_cost:1)
+		));
+		
+		run(str('/data modify entity %s Offers.Recipes append value %s', cleric~'command_name', trade));
```
Ghoulboy78, line 117: Doesn't scarpet have like `modify('nbt_merge')` or smth? I know I used to use run() as well as while back before gnembon added it, but the native scarpet is faster.

opsaaaaa:
Thanks for the review, ill start making some adjustment. 

Yes, Those Potions where chosen because they are not available as potion form in survival. You can attain many of the effects in other ways tho. 

I think [this line](https://github.com/gnembon/fabric-carpet/blob/master/docs/scarpet/api/Entities.md#entity-modification) in the scarpet api documentation is outdated.
>  Currently there is no ability to modify NBT directly, but you could always use run('data modify entity ...'). 
I am glad cause the `modify` method is a lot better.

opsaaaaa:
I have made the requested changes.

Ghoulboy78 (approved):
Yeah, looks pretty good

Firigion (commented):
Looks good to me, also works good. Why no Luck potion, tho?

opsaaaaa:
@Firigion.
I think I simply forgot to. I remember planning on it. Ill update it when I have some time.

opsaaaaa:
@Firigion I added luck potions to the available trades and updated the description

Ghoulboy78 (approved):
Looks pretty good

programs/survival/potion_master_cleric.sc
```diff
@@ -0,0 +1,169 @@
...
+	// Dont add any more potions to a potion master
+	modify(cleric, 'tag', 'IsPotionMaster'); 
+
+);
+
+__on_player_interacts_with_entity(player, entity, hand) -> (
+	if(entity~'type'=='villager',
+		nbt = entity~'nbt';
+		// If Villager is a cleric has a high enough level and is not already a potion master
+		if(nbt:'VillagerData':'profession'=='minecraft:cleric' &&
```
Firigion, line 160: You could make this line into `entity~'name' == 'cleric'` if you wanted to, but this works too.

programs/survival/potion_master_cleric.sc
```diff
@@ -0,0 +1,169 @@
...
+
+);
+
+__on_player_interacts_with_entity(player, entity, hand) -> (
+	if(entity~'type'=='villager',
+		nbt = entity~'nbt';
+		// If Villager is a cleric has a high enough level and is not already a potion master
+		if(nbt:'VillagerData':'profession'=='minecraft:cleric' && 
+			nbt:'VillagerData':'level' >= global_potion_master_level &&
+			parse_nbt(nbt:'Tags')~'IsPotionMaster' == null
```
Firigion, line 162: Comon practice is to give your tags a namespace specific to the app that ensures that no other app or datapck will colide with yours. For example, in this case I'd use something like `PotionMaster:IsPotionMaster` or if you want it less verbose `PMC:IsPotionMaster`.

Firigion (changes requested):
Just change the name of the ga and I'll approve. i didn't find any bugs, although I admit I only tested a dozen villagers.

programs/survival/potion_master_cleric.sc
opsaaaaa: That would be more readable. I think I prefer testing the profession directly. Plus I believe it is technically possible to rename villagers in survival.

opsaaaaa:
I am not sure what "ga" means in this context. I assumed it meant changing the IsPotionMaster tag. 

> Just change the name of the ga and I'll approve. i didn't find any bugs, although I admit I only tested a dozen villagers.

Firigion (approved):
Yup, I menat tag

PR #264, opened 2021-11-23
A fix for `playerme` hotbar

Desynq:
`playerme hotbar 0` no longer works with more recent versions of Carpet and `playerme hotbar 9` errors. The suggestions also hint at this mishap as well.

The integer range from the argument should instead range from 1 to 9 instead of 0 to 8.


altrisi (approved):
LGTM, @BisUmTo you may want to check it out/update the version in your repo

BisUmTo (approved):

PR #269, opened 2021-12-11
No Useless Llamas - New App

opsaaaaa:
A new app that maxes out a llamas inventory when a player clicks the llama with a chest.


Firigion (approved):
Yup, no coments for me. Worls flawlessly. You may want to add something to avoid changing the NBT of a lama you already modified. But it's not a critical issue at all, so I'll approve regardless, in case you dont' care.

opsaaaaa:
Humm. I think the use cases for the tag is rather small.

The only use case i can think of is a server owner wanting to set the inventory size of particular lamas to a lower size and didn't want the players increasing the size again by clicking them with a chest. But if that was the case I also doubt they would use this script in the first place.

After thinking about it, you are correct I don't wish to add the tag check feature.

rv3r (approved):
Simple, easy, and works well. Nice job. Only note is that first line of the script says "could a llama" unlike the description. Ignore this if you don't care about that grammar issue.

Issue #14, opened 2019-04-24
remove dat stupid thing where script prints result of each subexpression


gnembon:
I agree - its annoying

Issue #12, opened 2019-04-23
script run and script invoke... should return proper numerical result as a return value so that they work with /execute store


gnembon:
general workaround is to call run('execute store...) inside the script. This would allow for more custom ways to store different types of results. But I agree, it would be good it if worked with execute store

gnembon:
general workaround is to call run('execute store...) inside the script. This would allow for more custom ways to store different types of results. But I agree, it would be good it if worked with execute store

gnembon:
done

Issue #11, opened 2019-04-22
eval() - execute string contents.


bwucke:
eval() gives every scripting language a lot of power - ability to construct scripts, get them from unusual inputs and store in unusual ways. Examples: text of a written book. Items with custom behaviors encoded in the item metadata, input from external plugins e.g. discord chat, and so on. It's also probably the easiest way to implement lambda expressions, especially if paired with means to extract function source as a string.

encloinc:
an eval statement would be amazing!

gnembon:
lets assume we have some code written in a book. What would make it execute that code? some player action?
I am asking because with commandblocks its easy - pressing a button on a commandblock is essentially doing .eval() on it. How this would work on a book for example.

bwucke:
A command block still, or whatever other 'code execution environment', a chat window command, etc, wherever the eval() is used - it would be just another function. Using a command block would require a tiny snippet of code selecting the container of the command text (e.g. the book), and potentially condition when to execute, like place in a named item frame or in the off-hand (unless the command block triggers on a button press / redstone input), then pass the extracted code to eval. 

And yes, pressing a button on a command block is essentially doing .eval() _on it_ - that's why I believe the functionality should be pretty easy to implement, simply expose that very call to general scripting with the input parameter being the function's parameter instead of that one specific block entity data field. 

I'm even fairly sure one could even cheese it currently as a function in carpet script that saves a block in the world, summons a command block in its place setting its command to the function's parameter, activates the command block, then despawns it and restores the block it had replaced. A little messy with all that unnecessary block spawning and despawning but it would work. Doing it elegantly through a proper call without spawning unnecessary stuff into the world would be the big contribution of eval().

PegasusEpsilon:
`/script run run('script run ' + evaluated_string)`

Issue #10, opened 2019-04-22
loaded() always returns 1.


bolli24:
Even used on blocks in chunks that were never generated before the loaded() function returns 1.

gnembon:
Yeah, noticed that as well - OmegaLOL - by accident in reformatting I made all calls to loaded and loaded_ep - load that block. Its already fixed in current running master on carpetmod. Will release in some time as a public release, but for now you would need to wait, or build yourself form the carpetmod repo.

gnembon:
fixed on carpetmod master

bolli24:
Great to hear that. Thanks!

PR #270, opened 2021-12-13
Adding tag and block support to the filters

Firigion:
Adding a few more features ot `block_counter.sc` to make it more flexible and easy to use. I'll start the PR, but will prolly be adding to it as I go, since idk how many features I wanna implement. This is the list i got:

[*] tag filtering
[*] block name suggestions for filters
[] multiple blocks or filters (prolly not gonna do it)
[] print to file
[] save `y` value sof blocks to tally per  value isntead of total
[] automatic histograms once I add y values

### DIFF
--- a/programs/utilities/block_counter.sc
+++ b/programs/utilities/block_counter.sc
@@ -1,9 +1,15 @@
+//Counts blocks in a given region, optionally filtering by block type or tag. 
+//Prints the output into a chat or a file.
+//By gnembom and Firigion
+
 __config() ->
 {
    'commands' -> {
       '<from_pos> <to_pos>' -> 'count_blocks',
       '<from_pos> <to_pos> <filter>' -> 'count_blocks_filter',
-   }
+      '<from_pos> <to_pos> <blockpredicate>' -> 'count_blocks_predicate',
+   },
+   'allow_command_conflicts' -> true
 };
 
 count_blocks(from_pos, to_pos) ->
@@ -18,6 +24,22 @@ count_blocks_filter(from_pos, to_pos, filtr) ->
     stats = {}; total = 0;
     volume(from_pos, to_pos, total += 1; if (_~filtr, stats:str(_) += 1 ) );
     tally(stats, total);
+);
+
+count_blocks_tag(from_pos, to_pos, tag) ->
+(
+    stats = {}; total = 0;
+    volume(from_pos, to_pos, total += 1; if (block_tags(_, tag), stats:str(_) += 1 ) );
+    tally(stats, total);
+);
+
+count_blocks_predicate(from_pos, to_pos, predicate) ->
+(
+    [block, tag, trash, trash] = predicate;
+    if(block==null, 
+      count_blocks_tag(from_pos, to_pos, tag),
+      count_blocks_filter(from_pos, to_pos, filtr)
+    )
 );
 
 tally(stats, grand_total) ->
@@ -47,3 +69,4 @@ dpad(num, wid) ->
    if (length(strn) < wid, strn = '...'*(wid-length(strn))+strn);
    strn;
 )
+


Firigion:
it currently saves a json map. Maybe it should save a csv?

Firigion:
I'm splitting the code between the regular scan and the scan that separates reuslts per layer, because the latter is slower by something between 30% and 100% from my testing. Any suggestions on the matter are welcome.

The print out of the results in the case where scan is separated by layers is pretty jank right now, it just repeats the standard output for each layer once. I'm all open to suggestions regarding that too.

Firigion:
Oh, and I'm using the infamous ` 'allow_command_conflicts' -> true` in the config. I think this is one of the few cases where it makes sense.
The code is also clean't, but the repeated code makes for faster execution, which I think is more important than readability in a case where you need to likely scan multiple millions of blocks. I'll do some benchmarks later to see if it's worth keeping the multiple duplicate code lines.

Firigion:
This is a comment for whoever decides to review this:
I _could_ abstract the scan functions into a single `master_scan` function that's something like this
```
master_scan(from, to, condition) ->(
  stats = {}; total = 0;
  volume(from, to, if(call(condition, _), stats:str(_) += 1 ) ; total += 1; );
  tally(stats, total);
);
```
And then have the three scan functions reduced into
```
count_blocks(from_pos, to_pos, tag) ->(
  master_scan(from_pos, to_pos, _(b) -> true )
);
count_blocks_tag(from_pos, to_pos, tag) ->(
  master_scan(from_pos, to_pos, _(b, outer(tag)) -> block_tags(b, tag) )
);
count_blocks_filter(from_pos, to_pos, filter) ->(
  master_scan(from_pos, to_pos, _(b, outer(filter)) -> b~filter )
);
```
That would reduce around 40 lines of code to 10 and make it more readable and maintainable. However, doing this also slows down the scan by about 30 to 50%, depending on the type of scan, so I'm choosing to leave the current version in.

This fact, added to what I mentioned above
> I'm splitting the code between the regular scan and the scan that separates reuslts per layer, because the latter is slower by something between 30% and 100% from my testing. Any suggestions on the matter are welcome.

makes for a pretty nasty, but faster code.

CrayonEaterExtrodinare:
As per discord:
**#1** The reorganization of the `/block_counter` options is greatly appreciated

**#2** All functions reply with an understandable reply if the script (and it's options) are ran prior to a scan of an area.

**Testing command/area: `/block_counter -512 -64 -512 511 319 511 diamond` *Because I'm confident in what that ore-distribution is supposed to look like, and the area is 4 regions* **

**#3** Result of initial test took 7 min 25 seconds *calculated by identifying time stamps in the Multimc Minecraft log* This isn't a critique or suggestion, it's a datapoint.

**#4** `autosave true` functioned as intended

**#5** `10.json` was appropriately named sequential to previous saves in the `scripts` folder. I.E. There was already an `9.json` in said folder

**#6** The save omits, properly, layers void of specified blocks

**#7** Suggestion: In the saved file, there is no record/note of the following: *this meaning - the only way to identify, what parameters were used to create the save file is to be present at the scripts execution (as the parameters also do not appear in the Multimc Minecraft Log*
•What area was searched? (Suggest addition of detail at the top of `#.json` for ease of understanding "The data below is a record of..."
•What filter was applied? (Ex. "ore"? "diamond"? "diam"? "mond"?
•`Scanned area:` This result is present in game, via the chat's *Click to display results for  #*, This detail is also not recorded in the *Multimc Minecraft Log* (Suggest addition of detail at the top of `#.json` for ease of understanding.)
•`Total tally:` This result is present in game, via the chat's *Click to display result for #*, This detail is also not recorded in the *Multimc Minecraft Log*  (Suggest addition of detail ***IF*** the search returned more than 1 block type in the results. Ex. "deepslate_diamond" will only have 1 block type in the results for each layer. "diamond" results could include 3 different types.)

*That last point, evokes memories of being handed a graph - without labels on the "vertical" and "horizontal" lines*

**#8** The in chat selection works ***ASTOUNDINGLY*** Easy to understand and navigate.

**#9** `book` ...eagerly awaited

**#10** `histogram` displays data accurately (able to be understood)

**#11** `histogram` orders data counter intuitive to reader assumption. Ex. Negative `Y=` are at the top, while positive at at the bottom, meaning the user has to mentally 'flip' the image in their head to get a accurate representation of how things appear in world. *NOTE: Users can be trained to understand this, know that the initial assumption didn't account for this and the ore distribution appeared inverse.*

**#11** `histogram` selector is able to be understood. I know that #1 A scan must be ran prior, and #2 an individual block must be specified

**#12**  `last` selector works **PERFECTLY** I have no fear of losing the chat/having to repeat a scan.

**#13** `tally_level` Works as intended, and repeats data 1:1

**#14** `tally_level` Unsure as to why this selector suggests 3 numerical values at the end? `-123` `0` `123` 

**#14** `tally_level` shows `[<from_pos>] <to_pos> [<blockpredicate>|<filter>]` when only a single `Y=` value is required.

Firigion:
Implemented comments `#11` and `#7`. Also found a bug where filtering by block type didn't work correctly, so I had to make yet another scan function. I hope these don't keep piling up.

CrayonEaterExtrodinare:
#1 Suggested arguments for all options are `[<from_pos>] <to_pos> [<blockpredicate>|<filter>]`. These aren't applicable to the following options:
•autosave
•book
•histogram
•last
•layed_scan
•save
•tally_level

Everything else appears to be in working order, and able to be understood. :)

programs/utilities/block_counter.sc
```diff
@@ -46,4 +165,180 @@ dpad(num, wid) ->
...
+         str('%s [%s] ', color, _), 
+         '^g Click to display results for '+_,
+         str('!/%s %s %s', system_info('app_name'), fun, _)
+      ];
+      put(output, null, line, 'extend');
+   );
+   output
+);
+
+histogram(block) -> (
```
rv3r, line 216: Finer histogram code if you want to put it in
```
histogram(block) -> (
   block = str(block);
   blocks = global_data:'blocks';
   block_characters = ['','▏','▎','▍','▌','▋','▊','▉','█'];
   divisions = length(block_characters) - 1;

   if(blocks:block==null, _error(block + ' not available from last scan.'));
   maxcount = 0;
   values = map(blocks:block,
      y = _;
      count = global_data:'stats':y:block;
      maxcount = max(maxcount, count);
      [y, count];
   );
   divisor = ceil(maxcount/30); //to make sure the histogram bar is not longer than the chat box
   print('');
   print(format('y Results for ', 'yb '+block));
   for(values, 
      [y, count] = _;
      ratio = count / divisor;
      print(format(str('w %s %s%s', hpad(y, 4), floor(ratio) * block_characters:(-1),block_characters:round(divisions * (ratio % 1))), '^g count: '+count) )
   );
);
```

programs/utilities/block_counter.sc
```diff
@@ -1,28 +1,147 @@
+//Counts blocks in a given region, optionally filtering by block type or tag. 
+//Prints the output into a chat or a file.
+//If vertical scan is set to true, it will return a separate report for each y value.
+//By gnembom and Firigion
```
rv3r, line 4: Misspelled gnomgnom

Firigion: my bad

programs/utilities/block_counter.sc
Firigion: Not sure which one I prefer. What do people think? New histogram is more precise, but idk if that precision is needed or useful. You get the exact value if you hover over the bar, btw.

Current look of histogram is
image

and suggested new look (with better precision) is
image

rv3r: Just for reference, ｜(fullwidth vertical line, \uFF5C) is thinner than | (vertical line or pipe, \u007C) and offers even more resolution than █ (full block \u2588)

programs/utilities/block_counter.sc
altrisi:
```suggestion
//By gnembon and Firigion
```

Firigion: Oh, I thought it was a joke. sorry, i really don't see that sort of typos.

programs/utilities/block_counter.sc
Ghoulboy78: hmm. Personally I prefer rv3r's solution

programs/utilities/block_counter.sc
```diff
@@ -1,28 +1,151 @@
...
+    stats = global_data:'stats'; //save reference for easier access
+
+    if(global_vertscan,
+      // do a scan separating results by y value
+      area = global_data:'area';
+      [miny, maxy] = get_minmax_pos(from_pos, to_pos);
+      total = 0;
+
+      volume(from_pos, to_pos, 
+         stats:(miny+floor(total/area)):str(_) += 1; //very slightly faster than calling pos(_):1
```
Ghoulboy78, line 42: if you don't wannt call `pos(_):1` can't you just call `_y`?

programs/utilities/block_counter.sc
```diff
@@ -1,28 +1,151 @@
...
+    stats = global_data:'stats'; //save reference for easier access
+
+    if(global_vertscan,
+      // do a scan separating results by y value
+      area = global_data:'area';
+      [miny, maxy] = get_minmax_pos(from_pos, to_pos);
+      total = 0;
+
+      volume(from_pos, to_pos, 
+         if (_~filtr, stats:(miny+floor(total/area)):str(_) += 1); //very slightly faster than calling pos(_):1
```
Ghoulboy78, line 68: again with `_y` here

programs/utilities/block_counter.sc
```diff
@@ -46,4 +169,218 @@ dpad(num, wid) ->
...
+   print(format('gi No blocks were found with requested filter ' + global_data:'filter'));
+   exit();
+);
+
+/////////////////////////////
+///// Others
+/////////////////////////////
+
+global_data = {};
+make_data_stuct(from, to, filter) -> (
```
Ghoulboy78, line 310: struct, maybe?

programs/utilities/block_counter.sc
```diff
@@ -1,28 +1,151 @@
...
+      volume(from_pos, to_pos, 
+         stats:(miny+floor(total/area)):str(_) += 1; //very slightly faster than calling pos(_):1
+         total += 1; 
+      );
+      vertical_report(),
+
+      // do a regular scan
+      volume(from_pos, to_pos, stats:str(_) += 1 );
+      tally(global_data:'stats', global_data:'volume');
+      global_data:'total' = sum(... values(global_data:'stats'));
```
Ghoulboy78, line 50: If you're counting all the blocks here, instead of summing all the items of the list couldn't you just do dx * dy * dx?
Cos you have `from_pos` and `to_pos`, so the total is: `abs((from_pos:0 - to_pos:0)*(from_pos:1 - to_pos:1)*(from_pos:2 - to_pos:2))`

programs/utilities/block_counter.sc
```diff
@@ -1,28 +1,151 @@
...
+
+      volume(from_pos, to_pos, 
+         if (_~filtr, stats:(miny+floor(total/area)):str(_) += 1); //very slightly faster than calling pos(_):1
+         total += 1; 
+      );
+      vertical_report(),
+      // do a regular scan
+      volume(from_pos, to_pos, total += 1; if (_~filtr, stats:str(_) += 1 ) );
+      tally(global_data:'stats', global_data:'volume');
+      global_data:'total' = sum(... values(global_data:'stats'));
```
Ghoulboy78, line 75: wait, I may have gotten lost in the code here, but why not just set this to be the `total `variable that you already have? ik that you seem to be incrementing the total each time instead of only when the block matches the filter, but u also don't seem to be doing much with that variable (unless I'm mistaken), so you could just use that no? It would be faster than unpacking and summing after you've done your calculations

PR #274, opened 2021-12-17
Added block collision box library

rv3r:
This library provides the solid collision box of every block and block state combination in the game, or `null` if a block does not have solid collision.
Bounds are provided as list of pairs of triples representing vertices of the rectangular prisms that comprise collision box if block existed at [ 0, 0, 0 ].
```
__bounds( 'flowering_azalea' ) ->

[
  [
    [ 0.375, 0, 0.375 ],
    [ 0.625, 0.5, 0.625 ]
  ] ,
  [
    [ 0, 0.5, 0 ],
    [ 1, 1, 1 ]
  ]
]
```
```
__bounds( 'spruce_fence_gate[open=true]' ) -> null
```
```
__bounds( 'stone' ) ->

[
  [
    [ 0, 0, 0 ],
    [ 1, 1, 1 ]
  ]
]
```
Bounds function can take block(`block( 67, 10, -92 )`), triple(`[ 995, 23, -37 ]`), or string(`'stone'` or `'hopper[facing=east]'`). Can also provide the following:
- boolean representing if point is in a block
```
__insideblock( [ 106.3718, 72, -447.2 ] ) -> false
```
- filtered and sorted list of blocks
```
__filter_sort_direction( [ 'iron_bars', 'cake[bites=3]', 'grindstone', [ -54, 0, -199 ] ], '-x', 'min', 'exclude', 0) ->
  [ iron_bars, cake, grindstone ]
```
- all(tm) unique collision boxes that exist
```
example return value too big to fit in this margin but it's just a list of 573 blocks that's mostly cobblestone walls and chorus plants
```
- drawn bounds of a block with any color and fill
```
__draw_bounds( [ -69, 1, -186 ], 0xFF000FF, 0x0000FF55 ) ->
```
bounds

- list of blocks an entity is currently colliding with
```
__entity_block_collision( player('rv3r') ) -> [ lily_pad, comparator, piston_head ]
```

For a fun time, run
```
/script run import('collision','__entity_block_collision','__draw_bounds');__on_tick() -> ( for( __entity_block_collision(player()), __draw_bounds(_)))
```
and bump into a bunch of blocks

### **Todo list**
- [x] make this list look better
- [ ] can anything get sped up or are some functions always going to require a thread/be really slow?
- [ ] can iron bars and fences be combined?
- [ ] can candles, turtle eggs, and pickles be combined?
- [ ] maybe use early `return()` calls for blocks that don't have the final rotations/flips, might speed things up
- [ ] make sure all the debug prints and deprecated functions made it out
- [ ] don't think draw bounds checks which dimensions to draw in
- [ ] smh potted saplings, mushrooms, fungi, roots don't work right now, regex time!
- [ ] new 1.18 block tags might be handy
- [ ] can probably remove underscores from some function names
- [ ] code examples
  - [ ] `__bounds()`
  - [ ] `__draw_bounds()`
  - [ ] `__all_collision_blocks()`
  - [ ] `__filter_sort_direction()`
  - [ ] `__entity_block_collision()`
### Feedback wanted

1. Should this just be a default scarpet feature instead of a library?
2. I need better names for some functions(this means **you** `__filter_sort_direction()`).
3. Honestly that whole function might need to be redone
4. Should I add function to offset bounds to position in world? eg `[[[102,30,-10][103,31,-9]]]` instead of `[[[0,0,0],[1,1,1]]]`
5. Should double blocks(doors, pistons, etc) have both hitboxes or only one?
6. Hopper facing relies on list order being consistent. Is this robust enough?
7. Bamboo, pointed dripstone, open shulker boxes, etc have inconsistent collision boxes that can't be determined with `block_state()`. I'm currently providing a solid block for shulker boxes and the maximum possible collision box for others. How should these be handled?
8. Is adding more boxes good enough or should the originals be stretched? eg double chest, some iron bars and fences, etc
9. Better comments needed anywhere?
10. Are null hitboxes for fluids and portals okay?
11. I haven't manually checked every block for errors or correct hitboxes. Please complain if something is wrong.

Feel free to ask questions and provide feedback!

### DIFF
--- a/programs/libraries/collision.scl
+++ b/programs/libraries/collision.scl
@@ -1,0 +1,1357 @@
+//default direction for each block when placed with /setblock and no additional properties
+global_default_direction = {
+	['piston$','north']
+	['^piston_head$','north'],
+	['^small_amethyst_bud$','up'],
+	['^medium_amethyst_bud$','up'],
+	['^large_amethyst_bud$','up'],
+	['^amethyst_cluster$','up']
+};
+
+//blocks that are transparent but still have a full block collision box
+global_full_transparent = [
+	'leaves$',
+	'glass$',
+	'ice$',
+	'glowstone',
+	'beacon',
+	'redstone_block',
+	'sea_lantern',
+	'observer',
+	'tnt',
+	'chorus_flower',
+	'shulker_box$'
+];
+
+//blocks that pass solid() but are not full blocks
+global_not_solid = [
+	'soul_sand'
+];
+
+//list of tags of blocks with no collision box
+global_no_box_tags = [
+	'small_flowers',
+	'tall_flowers',
+	'banners',
+	'rails',
+	'buttons',
+	'pressure_plates',
+	'signs',
+	'bee_growables',
+	'corals',
+	'wall_corals',
+	//just gonna leave portals here for now unless someone needs them
+	'portals'
+];
+
+//then we go to names of things with no collision box
+//makes the search a little bit faster
+global_no_box_blocks = [
+	'air$',
+	'_stem$',
+	'torch$',
+	'redstone_wire',
+	'mushroom$',
+	'nether_wart',
+	'^tripwire',
+	'fern$',
+	'lever',
+	'sugar_cane',
+	'glow_lichen',
+	'grass$',
+	'fire$',
+	'^dead_',
+	'_fungus$',
+	'_roots$',
+	'^kelp',
+	'^nether_sprouts$',
+	'^spore_blossom$',
+	'^scaffolding$',
+	'cobweb',
+	'small_dripleaf',
+	'_sapling$',
+	'vine',
+	'powder_snow',
+	'^light$',
+	'structure_void',
+
+	//for now, we'll set these liquids as null
+	'water',
+	'lava',
+	'bubble_column'
+];
+
+//map of all weird collision boxes
+//then we do the basic, default collision box of everything else
+global_box_map = {
+	['_slab$',
+		[
+			[[0,0,0],[1,0.5,1]]
+		]
+	],
+	['_stairs$',
+		[
+			[[0,0,0],[1,0.5,1]]
+		]
+	],
+	['^potted_',
+		[
+			[[0.3125,0,0.3125],[0.6875,0.375,0.6875]]
+		]
+	],
+	['^flower_pot$',
+		[
+			[[0.3125,0,0.3125],[0.6875,0.375,0.6875]]
+		]
+	],
+	['^end_portal_frame$',
+		[
+			[[0,0,0],[1,0.8125,1]]
+		]
+	],
+	['^brewing_stand$',
+		[
+			[[0.0625,0,0.0625],[0.9375,0.125,0.9375]],
+			[[0.4375,0.125,0.4375],[0.5625,0.875,0.5625]]
+		]
+	],
+	['^enchanting_table$',
+		[
+			[[0,0,0],[1,0.75,1]]
+		]
+	],
+	['^farmland$',
+		[
+			[[0,0,0],[1,0.9375,1]]
+		]
+	],
+	['^ender_chest$',
+		[
+			[[0.0625,0,0.0625],[0.9375,0.875,0.9375]]
+		]
+	],
+	['^dirt_path$',
+		[
+			[[0,0,0],[1,0.9375,1]]
+		]
+	],
+	['anvil$',
+		[
+			[[0.125,0,0.125],[0.875,0.25,0.875]],
+			[[0.25,0.25,0.1875],[0.75,0.3125,0.8125]],
+			[[0.375,0.3125,0.25],[0.625,0.625,0.75]],
+			[[0.1875,0.625,0],[0.8125,1,1]]
+		]
+	],
+	['_bed$',
+		[
+			[[0,0,0.8125],[0.1875,0.1875,1]],
+			[[0.8125,0,0.8125],[1,0.1875,1]],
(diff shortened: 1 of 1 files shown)


ch-yx:
some block`s shape depends on who you are. such as snower snow and scaffolding.🤷‍♂️

programs/libraries/collision.scl
```diff
@@ -0,0 +1,1356 @@
...
+		]
+	],
+	['campfire$',
+		[
+			[[0,0,0],[1,0.4375,1]]
+		]
+	],
+	['^honey_block$',
+		[
+			[[0.0625,0.0625,0.0625],[0.9375,0.9375,0.9375]]
```
ch-yx, line 232: wrong. the second 0.0625 should be 0.

ch-yx:
random offset of bamboos like blocks is based on its x and z.
```
    public static long hashCode(int x, int z) {
        long l = (long)(x * 3129871) ^ (long)z * 116129781L;
        l = l * l * 42317861L + l * 11L;
        return l >> 16;
    }
```

then, the result of the hash function is used to calc the offset.
           `offsetX=(((hash & 15) / 15) - 0.5) * 0.5`
           `offsetZ=(((hash >> 8 & 15) / 15) - 0.5) * 0.5`
then it is clamped between +- maxModeloffset.
dripstone`s maxModeloffset is 0.125
others is 0.25

Ghoulboy78:
Just for implementation, in scarpet that would be:
```java
hashCode(x, z)->(
    l = bitwise_xor(x*3129871, z)*116129781;
    l = l * l * 42317861 + l * 11;
    bitwise_shift_right(l, 16);
)
```
I'm calling the function `hashCode` cos scarpet already has a native `hash_code` function which gets the has h of an object.

rv3r:
> some block`s shape depends on who you are. such as snower snow and scaffolding.🤷‍♂️

The script is meant mostly for projectiles, then players, so I'm not too concerned with getting entity states and such.

ch-yx:
> > some block`s shape depends on who you are. such as snower snow and scaffolding.🤷‍♂️
> 
> The script is meant mostly for projectiles, then players, so I'm not too concerned with getting entity states and such.

scaffolding s shape depends on whether your foot is higher than it, or whether you are shifting.........

rv3r:
> > > some block`s shape depends on who you are. such as snower snow and scaffolding.🤷‍♂️
> > 
> > 
> > The script is meant mostly for projectiles, then players, so I'm not too concerned with getting entity states and such.
> 
> scaffolding s shape depends on whether your foot is higher than it, or whether you are shifting.........

Clarification: This is mostly meant for reference and ideally runs in a vacuum. I don't want to query entities at all. I'd prefer the result of a `bounds()` call to be based solely on block state and nothing more, mainly for consistency.

rv3r:
> random offset of bamboos like blocks is based on its x and z.
> 
> ```
>     public static long hashCode(int x, int z) {
>         long l = (long)(x * 3129871) ^ (long)z * 116129781L;
>         l = l * l * 42317861L + l * 11L;
>         return l >> 16;
>     }
> ```
> 
> then, the result of the hash function is used to calc the offset. `offsetX=(((hash & 15) / 15) - 0.5) * 0.5` `offsetZ=(((hash >> 8 & 15) / 15) - 0.5) * 0.5` then it is clamped between ± maxModeloffset. dripstone`s maxModeloffset is 0.125 others is 0.25

@Ghoulboy78 

```
__hashCode(x, z)->(
	l = bitwise_xor(x * 3129871, z) * 116129781;
	l = l * l * 42317861 + l * 11;
	l = bitwise_shift_right(l, 16);
	[ ( bitwise_and( l, 15 ) / 15 - 0.5 ) * 0.5, ( bitwise_and( bitwise_shift_right( l, 8 ), 15 ) - 0.5 ) * 0.5 ]
);

__clamp(list, min, max) ->
(
	output = copy(list);
	for(output,
		if(_ > max,
			output:_i = max,
			_ < min,
			output:_i = min
		)
	);
	output
);

offset = __clamp(__hashCode(x,z),-0.25,0.25);
```
Does this code seem right?
The results aren't great...
bamboo
dripstone

rv3r:
Probably has something to do with the type I'm passing vs what the function ought to be given. Might have to do some manual recasting

ch-yx:
image_149
my script works fine here
[bamboo.sc](https://github.com/gnembon/scarpet/files/7776697/bamboo.txt)
.

rv3r:
Got it working. Thanks @ch-yx

Firigion:
Mark as draft while you work on it?

Firigion:
> make sure all the debug prints and deprecated functions made it out

you can have a wrapper around print galled `debug_print` or `dprint` or whatever that's basically
```
dprint(msg) -> if(global_debug, print(player(), msg);
```

and have a variable `global_debug` somewhere that you turn on while debugging. That way it's easier to go back to debug mode if you find an issue and also less work ahving to remove all the print statements.

rv3r:
Known issue: blocks that are positioned based on x-z location(bamboo, pointed dripstone) can sometimes fail collision checks because entities are touching but are not numerically in contact. Error only seems to occur in positive x and z directions. It appears that either both fail or neither fail, and this can change on reload.
Bamboo error:
  +x :   2nm
  +z : 12nm
Pointed dripstone error:
  +x : 4.9nm
  +z : 10nm

rv3r:
After further investigation, the above issue is likely related to the conversion between data types (float precision error or something like that). If someone can check the precision of the game's collision code or any rounding that occurs for bamboo/dripstone locations then I can just round values and hopefully that should solve this.

PR #275, opened 2021-12-18
Added maximum flight distance to cam app

Firigion:
I added a maximum range you can fly from your orignial position. Someone on discord asked for it and I though it was a decent suggestion to add.

Modifyin the global value `global_max_flight_range` to anything that evaluates true for `>0` will make it so you own't be able to flight further than the set value form your initial position. It's off by default. I think I covered all the edge cases with reloading and such.

### DIFF
--- a/programs/survival/cam.sc
+++ b/programs/survival/cam.sc
@@ -5,7 +5,11 @@
 // places spectating players without config in a safe places
 // and adds timeout for the player so they cannot switch instantly
 // change setting below to change that behaviour (0 to disable timeout)
-global_survival_timeout = 3;
+global_survival_timeout = 0;
+
+// optionally, you can restrict a player from flying too far from their original 
+// position (null or any non positive number to disable)
+global_max_flight_range = 10;
 
 // additional player checks are here. comment lines that you think are not needed
 __assert_player_can_cam_out(player) ->
@@ -20,6 +24,7 @@ __assert_player_can_cam_out(player) ->
 
 
 
+import('math', '_euclidean', '_vec_length');
 
 __config() -> {
     'stay_loaded' -> 'true'
@@ -146,6 +151,7 @@ __turn_to_camera_mode(player) ->
    modify(player, 'effect', 'night_vision', 999999, 0, false, false);
    modify(player, 'effect', 'conduit_power', 999999, 0, false, false);
    modify(player, 'gamemode', 'spectator');
+   __restrict_flight(player);
 );
 
 __survival_defaults(player) ->
@@ -163,4 +169,36 @@ __survival_defaults(player) ->
    );
    print(format('rb Cannot find a safe spot to land within 32 blocks.'));
    false;
+);
+
+__restrict_flight(player) -> 
+(
+   if(global_max_flight_range > 0, 
+      config = __get_player_stored_takeoff_params(player~'name');
+      if(!config, config = player~'pos'); // gotta restrict it somehow if no config was saved
+      start_pos = config:'pos';
+      __restrict_flight_tick(player, start_pos)
+   )
+);
+
+__restrict_flight_tick(player, start_pos) -> 
+(
+   if(player~'gamemode' != 'spectator', exit());
+
+   pp = player~'pos';
+   if(_euclidean(pp, start_pos)>global_max_flight_range,
+      direction = pp-start_pos;
+      direction = direction/_vec_length(direction);
+      new_pos = direction * global_max_flight_range + start_pos;
+      modify( player, 'pos', new_pos)
+   );
+
+   schedule(1, '__restrict_flight_tick', player, start_pos);
+);
+
+__on_player_connects(player) -> 
+(
+   if(player~'gamemode'=='spectator' &&  __get_player_stored_takeoff_params(player~'name');,
+      __restrict_flight(player)
+   )
 );


Firigion:
Well, I did forget an edge case: dimension changes.

Firigion:
Should I add a way to configurate if the distance limit is cylindrical or spherical? My thinking is 
```java
// optionally, you can restrict a player from flying too far from their original 
// position (null or any non positive number to disable)
global_max_flight_range = 10;
global_flight_restriction_cylindrical = false;
```
for the end-user-editable config.

programs/survival/cam.sc
```diff
@@ -164,3 +170,45 @@ __survival_defaults(player) ->
...
+      __restrict_flight_tick(player, start_pos, start_dim)
+   )
+);
+
+__restrict_flight_tick(player, start_pos, start_dim) -> 
+(
+   if(player~'gamemode' != 'spectator', exit());
+
+   pp = player~'pos';
+   if(_euclidean(pp, start_pos)>global_max_flight_range,
```
Ghoulboy78, line 195: U typically do `_euclidean_sq(pp, start_pos) > global_max_flight_range * global_max_flight_range` cos that way u avoid sqrt

programs/survival/cam.sc
```diff
@@ -5,7 +5,11 @@
 // places spectating players without config in a safe places
 // and adds timeout for the player so they cannot switch instantly
 // change setting below to change that behaviour (0 to disable timeout)
-global_survival_timeout = 3;
+global_survival_timeout = 0;
```
Ghoulboy78, line 8: Y changed to 0 here? Is it cos old default is 0?

programs/survival/cam.sc
```diff
@@ -5,7 +5,11 @@
...
+
+// optionally, you can restrict a player from flying too far from their original 
+// position (null or any non positive number to disable)
+global_max_flight_range = 10;
```
Ghoulboy78, line 12: Tbh most people are fine with /cam allowing you to move as far as you want, so maybe keep this off by default?

Ghoulboy78:
> Should I add a way to configurate if the distance limit is cylindrical or spherical? My thinking is
> 
> ```java
> // optionally, you can restrict a player from flying too far from their original 
> // position (null or any non positive number to disable)
> global_max_flight_range = 10;
> global_flight_restriction_cylindrical = false;
> ```
> 
> for the end-user-editable config.

I think it's a good idea, tho I think most people would stick to cylindrical, so make it default to true, and maybe also add option for rectangular, incase people want to allow exploration of rendered chunk? (with the spectatorsDontGenerateChunks option enabled, so that spectator mode works essentially like freecam). I mention that cos I've seen it done on some servers.

Ghoulboy78:
Yo firi, you gotta resolve conflicts between this and your other PR which just got merges, cos both use `__on_player_connects` in a similar way, so it's best you solve.

programs/survival/cam.sc
Firigion: absolutelyu, again. forgot to change defaults.

programs/survival/cam.sc
Firigion: Yeah, it's just that I didn't want to store the squared max radius globally, and it's not that we arecalculating thousands of square roots every tick. I can remove it if it bothers you.

programs/survival/cam.sc
Firigion: Forgot to change back while testing. There was a discussion about changing it to zero tho in discord. MIght leave it like this if people agree.

programs/survival/cam.sc
Firigion: Well, not anymore, because I use the same distance value for both absolute value and euclidian distance.

Firigion:
Last question for suggestions: 

Right now, the restriction is implemented as follows: if the player approaches the maximum distance, they bounce back as if they hit a force field or something (suggestion by RorarkCats). However, since spectator players can TP, if the player's distances to the original position is more than 1.5 times the maximum range, they get TPd back to an allowed position. Question is: should an actionbar message appear when you bounce against the invisible boundary of max flight distance? And how about when you you get TPd? My opinion rn is no to first one, yes to second one.

Ghoulboy78:
I concur

programs/survival/cam.sc
```diff
@@ -161,11 +169,109 @@ __survival_defaults(player) ->
          )
       )
    );
-   print(format('rb Cannot find a safe spot to land within 32 blocks.'));
+   display_title(player, 'actionbar', format('r Cannot find a safe spot to land within 32 blocks'));
+   //print(format('rb Cannot find a safe spot to land within 32 blocks.'));
```
MeeniMc, line 173: Remove printf

programs/survival/cam.sc
```diff
@@ -161,11 +169,109 @@ __survival_defaults(player) ->
...
+(
+   if(player~'gamemode' != 'spectator', exit());
+
+   pp = player~'pos';
+   distance = __get_distance(copy(pp), copy(start_pos));
+   
+   if(distance>global_max_flight_range,
+      if(distance>global_max_flight_range*1.5,
+         new_pos = __get_restriction_new_pos(pp, start_pos);
+         print(new_pos);
```
MeeniMc, line 203: remove print

MeeniMc (approved):
This works as intended
Left comments for two minor debug prints that should be removed

programs/survival/cam.sc
```diff
@@ -161,11 +169,109 @@ __survival_defaults(player) ->
...
+         modify( player, 'pos', new_pos); //in case they TP to another player
+         display_title(player, 'actionbar', format('y You are not allowed to go that far')),
```
MeeniMc, line 205: Maybe change the text to 'You cannot go in another dimension in cam mode'

programs/survival/cam.sc
```diff
@@ -161,11 +169,109 @@ __survival_defaults(player) ->
...
+         new_motion = __get_restriction_new_motion(pp, start_pos, player~'motion');
+         modify( player, 'motion', new_motion),
+      )
+      
+   );
+
+   if(player~'dimension' != start_dim,
+      run('execute in minecraft:'+ start_dim +' run tp @s ~ ~ ~');
+      modify( player, 'pos', start_pos);
+      display_title(player, 'actionbar', format('y You are not allowed to go that far'));
```
MeeniMc, line 216: Maybe change the text to 'You cannot go in another dimension in cam mode'

programs/survival/cam.sc
```diff
@@ -161,11 +169,107 @@ __survival_defaults(player) ->
...
+      display_title(player, 'actionbar', format('y You are not allowed to change dimensions'));
+   );
+
+   schedule(1, '__restrict_flight_tick', player, start_pos, start_dim);
+);
+
+__get_distance(pos1, pos2)->
+(
+   // messure flat distance for square and cylindrical
+   if (global_restriction_shape!=2,
```
Ghoulboy78, line 223: Is this meant to be `!=` or `==` for cylindrical distance?

Ghoulboy78 (approved):
Looks pretty good, just the one thing there which I'm not sure about

programs/survival/cam.sc
MeeniMc: Code is correct, if ==2 it is the spherical case, in which we should not collapse the 3d space on a 2d plane.

Firigion: This ^

MeeniMc:
Is there even enough devs with write access to have 2 reviews? :D 

This looks as ready as it can get to me.

JacktheDevel0per:
Maybe you can configure the box or whatever to move to that player you tp to?

Firigion:
I think that can be added, yes. But I think this app should be not very configuration dependent...
On second thought, idk how to detect a plyer TPing to another player using the spectator teleport tool.

PR #276, opened 2021-12-18
Fix for case where force-gamemode is on

Firigion:
When force-gamemode is on, the server puts a player that logged out in camera mode into survival mode when loggin in. this of course breaks things. A simple check should fix that.

### DIFF
--- a/programs/survival/cam.sc
+++ b/programs/survival/cam.sc
@@ -164,3 +164,5 @@ __survival_defaults(player) ->
    print(format('rb Cannot find a safe spot to land within 32 blocks.'));
    false;
 );
+
+__on_player_connects(player) -> if( __get_player_stored_takeoff_params(player~'name'), modify(player, 'gamemode' , 'spectator' ));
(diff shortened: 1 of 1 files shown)


programs/survival/cam.sc
```diff
@@ -164,3 +164,5 @@ __survival_defaults(player) ->
    print(format('rb Cannot find a safe spot to land within 32 blocks.'));
    false;
 );
+
+__on_player_connects(player) -> if( __get_player_stored_takeoff_params(player~'name'), modify(player, 'gamemode' , 'spectator' ));
```
altrisi, line 168: Maybe add a comment explaining why this exists, it's not easy to understand otherwise.

altrisi (approved):

programs/survival/cam.sc
Firigion: True

Ghoulboy78 (approved):

PR #279, opened 2022-01-07
Add experience bottling through enchanting table

Firigion:
Shift right-click on an enchanting table with an empty glass bottle and get a bottle o' enchanting. Average xp gain from bottles is 7, so I set the cost for bottling at 8.

### DIFF
--- a/programs/survival/enchanting_table_bottling.sc
+++ b/programs/survival/enchanting_table_bottling.sc
@@ -3,11 +3,11 @@ global_xp_cost = 8;
 __on_player_right_clicks_block(player, item, hand, block, f, h) -> (
 	if(
 		player~'gamemode_id'%2==0 && //survival or adventure mode
+		player~'sneaking' && //they didn't open the UI
 		item:0 == 'glass_bottle' &&
 		item:2 == null && //doesn't have nbt
 		block == 'enchanting_table' &&
-		player~'sneaking' && //they didn't open the UI
-		player~'xp' >= global_xp_cost, //they ahve enough xp
+		player~'xp' >= global_xp_cost, //they have enough xp
 
 		modify(player, 'add_xp', -global_xp_cost);
 		slot = if(hand=='mainhand', player~'selected_slot', -1);
--- a/programs/survival/enchanting_table_bottling.sc
+++ b/programs/survival/enchanting_table_bottling.sc
@@ -1,0 +1,18 @@
+global_xp_cost = 8;
+
+__on_player_right_clicks_block(player, item, hand, block, f, h) -> (
+	if(
+		player~'gamemode_id'%2==0 && //survival or adventure mode
+		item:0 == 'glass_bottle' &&
+		item:2 == null && //doesn't have nbt
+		block == 'enchanting_table' &&
+		player~'sneaking' && //they didn't open the UI
+		player~'xp' >= global_xp_cost, //they ahve enough xp
+
+		modify(player, 'add_xp', -global_xp_cost);
+		slot = if(hand=='mainhand', player~'selected_slot', -1);
+		inventory_set(player, slot, item:1 - 1);
+
+		run( str( 'give %s minecraft:experience_bottle', player))
+	)
+);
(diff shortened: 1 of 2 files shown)


opsaaaaa (approved):
I think this would work well in a survival context as is.
 
I found a bug related to setting your levels. However it might be a minecraft or scarpet bug.

If you set your levels or add level instead of xp then it stops working at a certain point.
```
/script run player()~'xp'   => 5
/experience add @p 1 levels/experience add @p 1 levels
/script run player()~'xp'  => 5
/experience add @p 1 points
/script run player()~'xp'  => 6
/experience set @p 70 points
/script run player()~'xp' => 6
/script run player()~'xp_level' => 26
``` 

It looks like there are three methods of getting xp in scarpet.
```
query(e,'xp')
query(e,'xp_level')
query(e,'xp_progress')
```

Looking a the difference between xp and levels, i suspect fixing this bug would be overly complicated in the context of a script. I doubt players will encounter the issue in normal survival anyway.

Firigion:
Yes, I noticed that, but as you say, handling it from the script side would be a looot of pain. The downside is that it mighight break compatibility with some datapacks or stuff like that, but I'm not too worried. We'll see what other perople think.

Ghoulboy78:
> I think this would work well in a survival context as is.
> 
> I found a bug related to setting your levels. However it might be a minecraft or scarpet bug.
> 
> If you set your levels or add level instead of xp then it stops working at a certain point.
> 
> ```
> /script run player()~'xp'   => 5
> /experience add @p 1 levels/experience add @p 1 levels
> /script run player()~'xp'  => 5
> /experience add @p 1 points
> /script run player()~'xp'  => 6
> /experience set @p 70 points
> /script run player()~'xp' => 6
> /script run player()~'xp_level' => 26
> ```
> 
> Looking a the difference between xp and levels, i suspect fixing this bug would be overly complicated in the context of a script. I doubt players will encounter the issue in normal survival anyway.
@opsaaaaa 
If you notice a bug there then pls report it to fabric-carpet and I will fix it (cos it's my code that's broken there)

opsaaaaa:
@Ghoulboy78 #1270

Ghoulboy78 (approved):

programs/survival/enchanting_table_bottling.sc
```diff
@@ -0,0 +1,18 @@
+global_xp_cost = 8;
+
+__on_player_right_clicks_block(player, item, hand, block, f, h) -> (
+	if(
+		player~'gamemode_id'%2==0 && //survival or adventure mode
+		item:0 == 'glass_bottle' &&
+		item:2 == null && //doesn't have nbt
+		block == 'enchanting_table' &&
+		player~'sneaking' && //they didn't open the UI
+		player~'xp' >= global_xp_cost, //they ahve enough xp
```
altrisi, line 10: Suggestion for checking the sneaking condition first (or after game mode) as it seems both a very cheap check and one that'll easily cancel the rest of checks in many cases.

programs/survival/enchanting_table_bottling.sc
```diff
@@ -0,0 +1,18 @@
...
+
+		modify(player, 'add_xp', -global_xp_cost);
+		slot = if(hand=='mainhand', player~'selected_slot', -1);
+		inventory_set(player, slot, item:1 - 1);
+
+		run( str( 'give %s minecraft:experience_bottle', player))
```
altrisi, line 16: Doesn't scarpet have a native function for this??????? (please tell me it does)

altrisi (commented):
Looks good!

Just two small things.

programs/survival/enchanting_table_bottling.sc
opsaaaaa: inventory_set is the closest thing I have seen. I am guessing there currently is not.

CommandLeo: You can use `slot = inventory_find(p, 'air')`

Firigion: Not quite. Until a proper give fucntion is added either to the language or the default libraries (so we can get them with `resources` in the config, or however that work), I'll keep the give command.

programs/survival/enchanting_table_bottling.sc
Firigion: I thought that since the block and item had already been queried and saved into a variable by the event handler, those checks would be cheaper than the sneak one. Ut was just a guess, tho.

altrisi (approved):

PR #281, opened 2022-01-07
skip_night.sc was in the wrong directory

Firigion:
For some reason the app was in the root directory and no one noticed :P


Ghoulboy78 (approved):
Lol

rv3r (approved):
Weird error but nbd, easy fix

PR #284, opened 2022-01-08
world_map fixes for 1.18

### LINKED ISSUE #280: World Map breaks in 1.18
**Describe the bug**
biome() changed in 1.18 - depth, scale, top_material and under_material are gone, exceptions will be thrown.

biome sampling needs to be done higher up because at y=0 you get can get cave biomes.

**To Reproduce**
Load world_map in a 1.18.1 save, eat a berry.

**Environment**
 - Carpet version: 1.4.56
 - Minecraft version: 1.18.1

**Additional context**

https://github.com/random-j-farmer/scarpet/blob/master/programs/survival/world_map.sc
works for me in 1.18.1, but I assume I broke it for earlier Minecraft versions.

I don't really know enough about minecraft to make a clean fix.

Setting the biome on the world map is also broken, 
I suspect setting it a y=0 does not work anymore in the overworld.

random-j-farmer:
A relatively clean fix for the 1.18 breakage.

Fixes #280

Still works in 1.17.1, I did not test with older versions because I deleted my old mod folders at one point.

For the new biomes, I just copied configuration for existing, somewhat similiar biomes.

```
global_biome_data:'meadow' = global_biome_data:'plains';
global_biome_data:'grove' = global_biome_data:'taiga_hills';
global_biome_data:'snowy_slopes' = global_biome_data:'snowy_mountains';
global_biome_data:'jagged_peaks' = global_biome_data:'snowy_mountains';
global_biome_data:'stony_peaks' = global_biome_data:'mountains';
```

This does not add visual markers for the new features (Lush & Dripstone Caves, Geodes, ...).  I would like to have that,
but I also want to play the game this weekend ;)

### DIFF
--- a/programs/survival/world_map.sc
+++ b/programs/survival/world_map.sc
@@ -7,6 +7,7 @@ use:
  - eat your berry
  or
  - /world_map tp
+ - /world_map help
 ';
 
 __config() -> {
@@ -24,7 +25,12 @@ global_player_chunk_map = {};
 
 // use threads to seamlessly generate chunk data
 // slower than synced but no lag spikes
-global_parallel = true;
+//
+// in 1.18.1 things break if this is on
+// i think chunk loading is parallel now in the base game
+// and this interacts badly with world_map's attempt
+// to parallelize
+global_parallel = system_info('game_major_target') < 18;
 
 // generation or degeneration
 global_generate_forward = true;
@@ -32,6 +38,8 @@ global_generate_forward = true;
 // output timing and progress information
 global_debug = false;
 
+// biome changes with cave biomes are causing the trouble with 1.18
+global_has_cave_biomes = system_info('game_major_target') >= 18;
 
 // some common decorations
 global_decorations = {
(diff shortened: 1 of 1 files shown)


altrisi:
Note for future reviewers (incl. self): `biome(<biome>, 'category')` has been removed in 1.19, so to properly test this you need to run on 1.18(.2 quite sure would work too). 1.19 compat can be looked into later, it shouldn't block this.

BTW sorry about being super slow with PRs! Thanks for contributing even if it's not in yet!

rv3r (commented):
Errors mentioned in issue seem to be fixed. New biomes seem to work fine (as far as I can tell). I can approve if we don't care who does it and just want someone to do it.

I've never used this script, so I never knew this was (or maybe just is) the longest script in here by leaps and bounds. Additionally, I love it when my game pauses for an entire minute while generating the map. Maybe I should've tested this on my home computer...

Ghoulboy78:
It doesn't work in any version above 1.18.2, cos certain key features were removed. For higher versions it would likely have to be rewritten entirely.

rv3r:
> It doesn't work in any version above 1.18.2, cos certain key features were removed. For higher versions it would likely have to be rewritten entirely.

Yep I figured as much.

> 1.19 compat can be looked into later

PR #286, opened 2022-01-20
Add instamine deepslate script

manyrandomthings:


Ghoulboy78 (approved):
Code looks good, now add it to the Readme.md

programs/survival/instamine_deepslate.sc
```diff
@@ -0,0 +1,58 @@
+// allows instamining deepslate if the player has an efficiency 5 netherite pick and haste 2
+// works even if the player doesn't have carpet mod on their client
+
+__config() -> {
+  'stay_loaded' -> true,
```
BisUmTo, line 5: Default value is true

manyrandomthings: This script doesn't load automatically without this

altrisi: That shouldn't be the case, it's true by default since ages ago.

altrisi: Resolved in Discord (it happens because of a carpet bug where it defaults to false with no `__config`).

altrisi (approved):

PR #287, opened 2022-01-24
Postmortem Death Count app

opsaaaaa:
Adds a scoreboard death count to a server after players have already died.
When players connect it sets their scoreboard death count to their statistic death count


programs/survival/postmortem_death_count.sc
```diff
@@ -0,0 +1,22 @@
...
+// Fetches the server deaths statistic to use in the scoreboard death count. 
+
+// when players connect it sets their scoreboard death count to their statistic death count
+
+_initialize() -> (
+    scoreboard_add( 'Deaths', 'deathCount');
+    for( player('all'),
+        _update_player_death_count(_);
+    );
+);
```
BisUmTo, line 12: You can write the body of this function directly at the end of the file.

Ghoulboy78 (approved):

CommandLeo:
Shouldn't this be in `utilities`?

BisUmTo:
I thought that you only die in survival mode... 
It is a survival utility 😅

PR #290, opened 2022-02-05
Delete event test

altrisi:
Deletes the event_test app from the app store.

It's not really useful other than for testing the outcomes of events, and it'll always be outdated against the one bundled in Carpet, that should always be preferred.


Ghoulboy78 (approved):

Firigion (approved):
yup

PR #291, opened 2022-02-11
Lodestone chunk loader

opsaaaaa:
Lodestones are Loadstones aka chunk loaders.
When a player clicks on a lodestone while sneaking it marks that chunk to be force loaded
and if they click on it again the unmarks that chunk.
Breaking the lodestone also unmarks the chunk.

Warning, with this script enabled players can unmark chunks admins intend to stay loaded.
Only use this script on your server if you would trust your players with the `/forceload` command.


programs/survival/README.md
```diff
@@ -126,6 +126,18 @@ Various scripts that modify various game elements, often replicating popular mod
...
+
+### [lodestone_chunk_loader.sc](https://github.com/gnembon/scarpet/blob/master/programs/survival/lodestone_chunk_loader.sc)
+#### By Opsaaaaa
+	Lodestones are Loadstones aka chunk loaders.
+	When a player clicks on a lodestone while sneaking it marks that chunk to be force loaded
+	and if they click on it again the unmarks that chunk.
+	Breaking the loadstone also unmarks the chunk.
+	
+	Warning, with this script enabled players can unmark chunks admins intend to stay loaded.
+	Only use this script on your server if you would trust your players with the `/forceload` command.
```
altrisi, line 138: Going as far as "Warning!" seems a bit much. Maybe something like "Note that" would be better.
Carpet Extra also uses the same trick for flowerPotChunkLoading (adds to forceload), so I don't think it's that bad. Users don't even need to know the implementation, and are likely to not know what chunks are forceloaded anyway.

altrisi (approved):
Looks good!

For a next iteration if you want to be safe against exposing forceload you could instead save those chunks in some list and use Scarpet's ticket functions to load chunks (while they were not the best at their time, I'd assume they have improved with time, else you can always make a bug report)

programs/survival/lodestone_chunk_loader.sc
```diff
@@ -0,0 +1,56 @@
+// Lodestones are Loadstones aka chunk loaders.
+// When a player clicks on a lodestone while sneaking it marks that chunk to be force loaded
+// and if they click on it again the unmarks that chunk.
+// Breaking the loadstone also unmarks the chunk.
+
+// Warning this script is not very smart,
```
Ghoulboy78, line 6: Same thing as what altrisi said above

Ghoulboy78:
Also, if the script is not very smart, couldn't you make it smarter by saving the chunks that the player forceleoads to disk? so that you can see what chunks have been forceloaded using this app, and if a chunk is forceloaded that is not on your list it means that an admin forceloaded it, an if a chunk isn't forceloaded which was on your list means an admin unloaded it, so you remove it from the list.
Then you save the list to disk with `__on_close()` and load it back with `__on_start()`

Ghoulboy78:
Though don't do it if it seems like too much.

opsaaaaa:
I thought about keeping track in a list. In the end I decided that most people using this script probably don't mind if their players can basically use forceload. so I decided not to add the extra complexity and just let people know about it. If it becomes something people want then perhaps it will be worth adding at that time.

opsaaaaa:
I figured scarpet had a way to load chunks but i couldn't find it, i don't know how i missed that. 
Looks like the `add_chunk_ticket` loads chunks for a certain amount of time, so I would need to add a schedule loop every 300 ticks to keep it loaded. Im glad to know about that, but I think I am satisfied with the forceload solution at the moment.

programs/survival/lodestone_chunk_loader.sc
```diff
@@ -0,0 +1,53 @@
...
+// and if they click on it again the unmarks that chunk.
+// Breaking the loadstone also unmarks the chunk.
+
+// Note that with this script players can unmark forceload chunks set by admins.
+
+global_chunk_regex = '\\[[\\d,\\s]+\\]';
+
+_block_forceload_chunk(block, mode) -> (
+    pos = pos(block);
+    return( in_dimension(block, run(str('/forceload %s %d %d', mode, pos:0, pos:2 ))) );
```
Firigion, line 12: do you need this exlicit return statement? It doesn't hinder performance, but just checking.

Firigion (approved):
Works fine. I agree both with the previous reviews and with the answer: for a simple app, it's good enough, it doesn't need to keep track of the app-loaded chunks if you don't want it to.

programs/survival/lodestone_chunk_loader.sc
altrisi: That return statement is removed by the compiler, no point hurting readability if they like to have it.

opsaaaaa: sometimes my brain just wants to see the return.

PR #296, opened 2022-02-23
Updated auto_ligher.sc to new 1.18+ light level

Ghoulboy78:
I added it as a global variable so you can change based on the version you're playing on. Should I use system_info to check automatically though?
I also added a boolean for lighting the ground cos some people want that too.

### DIFF
--- a/programs/survival/auto_lighter.sc
+++ b/programs/survival/auto_lighter.sc
@@ -37,6 +37,10 @@ global_effect_radius = 128;
 
 global_survival=false;
 
+global_min_light_level = 1; //1 for 1.18+, 8 for 1.17 and earlier
+
+global_light_ground = false; //Whether or not we want to light the surface
+
 spread_torches(player, initial_gamemode) ->
 (
 	if (global_spread_love && player~'holds':0 == 'torch' && player ~'gamemode_id' == initial_gamemode,
@@ -47,7 +51,7 @@ spread_torches(player, initial_gamemode) ->
 		loop(4000,
 			lpos = cpos+l(rand(d), rand(d), rand(d)) - d/2;
 			if (__distance_sq(cpos, lpos) <= dd  
-					&& air(lpos) && light(lpos) < 8 && sky_light(lpos) < 8
+					&& air(lpos) && light(lpos) < global_min_light_level && (global_light_ground || sky_light(lpos) < global_min_light_level)
 					&& solid(pos_offset(lpos, 'down')),
 				if (is_survival && not_able_loose_torch(player),
 					//running out of torches as survival
--- a/programs/survival/auto_lighter.sc
+++ b/programs/survival/auto_lighter.sc
@@ -5,6 +5,14 @@
 
 ///right click on a torch while looking at nothing to toggle.
 
+//Here you can customize the app to your needs
+global_effect_radius = 128; //How far away you want to send torches
+
+global_min_light_level = __light_level_for_version(); //1 for 1.18+, 8 for 1.17 and earlier
+
+global_light_ground = true; //Whether or not we want to light the surface
+
+
 // stay loaded
 __config() -> (
    m(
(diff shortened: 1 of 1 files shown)


CommandLeo:
> Should I use system_info to check automatically though?

Yes, support for 1.17 (and 1.16) should always be implemented whenever it's easily doable.

programs/survival/auto_lighter.sc
```diff
@@ -37,6 +37,10 @@ global_effect_radius = 128;
 
 global_survival=false;
 
+global_min_light_level = if(system_info('game_major_target')>=18, 1, 8); //1 for 1.18+, 8 for 1.17 and earlier
+
+global_light_ground = false; //Whether or not we want to light the surface
```
altrisi, line 42: I'd set this to true by default, given it's very common that people want to.

programs/survival/auto_lighter.sc
```diff
@@ -37,6 +37,10 @@ global_effect_radius = 128;
 
 global_survival=false;
 
+global_min_light_level = if(system_info('game_major_target')>=18, 1, 8); //1 for 1.18+, 8 for 1.17 and earlier
```
altrisi, line 40: Maybe extract the logic to a function like get level for version, so if someone is looking at it it doesn't seem like they shouldn't touch it or it'll break.

altrisi (commented):
Looks good, other than this I'd put those near the top so they're easy to see for someone who doesn't know scarpet and with the comments above so people know what they do immediately.

programs/survival/auto_lighter.sc
Ghoulboy78: sure

altrisi (approved):

programs/survival/auto_lighter.sc
```diff
@@ -5,6 +5,14 @@
 
 ///right click on a torch while looking at nothing to toggle.
 
+//Here you can customize the app to your needs
+global_effect_radius = 128; //How far away you want to send torches
+
+global_min_light_level = __light_level_for_version(); //1 for 1.18+, 8 for 1.17 and earlier
+
+global_light_ground = true; //Whether or not we want to light the surface
```
Firigion, line 13: Thwn the app is first loaded, it doesn't yet know about the function, you need to wrap this into an `on_start` or similar if we want to keep it at the top.
```suggestion
__on_start() -> (
	global_effect_radius = 128; //How far away you want to send torches

	global_min_light_level = __light_level_for_version(); //1 for 1.18+, 8 for 1.17 and earlier

	global_light_ground = true; //Whether or not we want to light the surface
);
```

programs/survival/auto_lighter.sc
```diff
@@ -47,7 +55,7 @@ spread_torches(player, initial_gamemode) ->
 		loop(4000,
 			lpos = cpos+l(rand(d), rand(d), rand(d)) - d/2;
 			if (__distance_sq(cpos, lpos) <= dd  
-					&& air(lpos) && light(lpos) < 8 && sky_light(lpos) < 8
+					&& air(lpos) && light(lpos) < global_min_light_level && (global_light_ground || sky_light(lpos) < global_min_light_level)
```
Firigion, line 58: I think ths logic is slightly flawed :/

Firigion (dismissed):
This doesn't look right... And it's not lighting the surface for me (1.18.1)
image

programs/survival/auto_lighter.sc
Ghoulboy78: Hmm, it worked fine when I tested it...

Ghoulboy78: oh wait Ik what I did wrong. I forgot to invert `global_light_ground`. Because if it's true, then we want to check for the sky light thing, whereas now we do the opposite, which isn't what we want. I'd changed this on personal copy after realising, must have forgotten to transfer change here.

BisUmTo (approved):
Seams to work fine both in 1.17 and in 1.18

Firigion:
The preading now works. However, it still doesn't light up the surface od the world for me, and I have no idea why (I double checked the logic, it looks good to me?). Dod the other reviewers get their ground torched (:fire:)?

Ghoulboy78:
@Firigion That could be because you're running it during the daytime? Or maybe u forgot to turn that setting?

PR #299, opened 2022-02-25
Add support for trader llamas to `no_useless_llamas`

opsaaaaa:
Turns out llama's and trader_llama's are different entities types.


Firigion (approved):
yup, works

altrisi (approved):

PR #300, opened 2022-03-16
plains and meadows

baford25:
Would it be possible to add functionality for the plains and meadows flower gradients as well?

Building a flower farm in a plains biome with half in a tulip section is often preferable to a flower forest.


altrisi:
I think you wanted to make an issue instead of a pull request, a pull request is used when you have code you want to add, not for commenting bugs or suggesting ideas.

I'm closing this as invalid, feel free to open an issue for this though.

PR #301, opened 2022-03-18
Create Loft

Cfer98:
This script lets you make splines and do a sort of loft function with those splines. The result is a highly customizable surface that is first rendered and then can be pastet with the command paste.
(made by Cfer98)

### DIFF
--- a/programs/building/Loft
+++ b/programs/building/Loft
@@ -1,0 +1,356 @@
+__config() -> {
+    'stay_loaded'->true,
+    'commands'->{
+        'wand'->['wand'],
+        'paste <Block> <replace>'->['paste'],
+        'paste <Block>'->['paste',null],
+        'paste'->['paste','diamond_block',null],
+        'undo'->['undo']
+    },
+    'arguments'->{
+        'Block'->{'type'->'term', 'suggest'->[
+			'white_concrete','yellow_concrete','orange_concrete','purple_concrete',
+			'lime_concrete','red_concrete','green_concrete','magenta_concrete',
+			'brown_concrete','gray_concrete','light_gray_concrete','blue_concrete',
+			'light_blue_concrete','pink_concrete','cyan_concrete','black_concrete'
+			]
+		},
+        'replace'->{'type'->'term','suggest'->['air','water','lava']}
+    }
+};
+
+global_tool='phantom_membrane';
+global_block='diamond_block';
+global_replace=null;
+global_sets=[]; //array containing current information about the m splines created
+global_points=[]; //vector of maximum n points of the current spline that is being created by the player
+global_cset=[]; //grid of nxm points created based on global_sets by the function _normalise_set()
+global_a=[]; //grid of a coefficients needed for bezier surfaces
+global_b=[]; //grid of b coefficients needed for bezier surfaces
+global_c=[]; //grid of c coefficients needed for bezier surfaces
+global_affected_block=[]; //vector containing information about block that the script changed in the last paste operation
+global_history=[]; //vector containing information about block that the script changed in the last N paste operation
+
+wand()->(
+    print('the wand is:');
+    print(global_tool);
+);
+
+__on_player_breaks_block(player, block)->( //starts a new spline
+    item_tuple = query(player,'holds',hand='mainhand'):0;
+    if(item_tuple==global_tool,
+        schedule(0, _(outer(block)) -> set(pos(block), block));
+        _draw_surf(1);
+        global_points=[];
+        global_points+=pos(block);
+        global_sets:(length(global_sets))=global_points;
+        _normalise_set();
+        _draw_surf(10000)
+    );
+    if(item_tuple!=global_tool,
+        if(global_sets!=[],
+            _draw_surf(1);
+            print('operation cancelled');
+            global_sets=[];
+            global_cset=[];
+            global_points=[]
+        )
+    )
+);
+
+__on_player_right_clicks_block(player, item_tuple, hand, block, face, hitvec)->( //expands the spline adding new points
+    if((get(item_tuple,0)==global_tool)&&(hand=='mainhand')&&(global_points!=[]),
+        _draw_surf(1);
+        global_points+=pos(block);
+        delete(global_sets,-1);
+        global_sets:(length(global_sets))=global_points;
+        _normalise_set();
+        _draw_surf(10000)
+    );
+    if(((get(item_tuple,0)!=global_tool)&&(hand=='mainhand')),
+        if(global_sets!=[],
+            _draw_surf(1);
+            print('operation cancelled');
+            global_sets=[];
+            global_cset=[];
+            global_points=[]
+        )
+    )
+);
+
+//this function transform the global_sets in a rectangoular grid (m*n) creating the loft using the created splines
+//it defines global_a, global_b and global_c array of coefficient points needed for the defination of the surfaces
+_normalise_set()->( 
+    global_cset=[];
+    m=length(global_sets); //m is the number of splines contained in global_sets
+    n=length(global_sets:0); //n will be the max number of points found in one of the splines
+    c_for(i=1,i<m,i=i+1, 
+        if(length(global_sets:i)>n,
+            n=length(global_sets:i)
+        )
+    );
+    if(n<=1, //in case each spline is composed by only one point (so not really a spline in this case)
+        global_cset=global_sets;
+    );
+    if(n>1,
+        c_for(i=0,i<m,i=i+1, //calculate new_points, a vector of n points for each splines that has less than n points
+            if(n>length(global_sets:i),
+                new_points=[];
+                //in case spline had 1 point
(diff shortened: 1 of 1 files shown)


programs/loft.sc
```diff
@@ -0,0 +1,356 @@
...
+    },
+    'arguments'->{
+        'Block'->{'type'->'term', 'suggest'->[
+			'white_concrete','yellow_concrete','orange_concrete','purple_concrete',
+			'lime_concrete','red_concrete','green_concrete','magenta_concrete',
+			'brown_concrete','gray_concrete','light_gray_concrete','blue_concrete',
+			'light_blue_concrete','pink_concrete','cyan_concrete','black_concrete'
+			]
+		},
+        'replace'->{'type'->'term','suggest'->['air','water','lava']}
```
Firigion, line 18: Well, as suggested on discord, move these two over to an actual block type, so that the minecraft command parser can handle stuff like `dirp` suggest autocompletions for blocks that are not concrete.

programs/loft.sc
```diff
@@ -0,0 +1,356 @@
+__config() -> {
+    'stay_loaded'->true,
+    'commands'->{
+        'wand'->['wand'],
```
Firigion, line 4: the brackets here are not needed. you can ignore this comment tho, it's just an aesthetic thing.

programs/loft.sc
```diff
@@ -0,0 +1,356 @@
...
+__on_player_breaks_block(player, block)->( //starts a new spline
+    item_tuple = query(player,'holds',hand='mainhand'):0;
+    if(item_tuple==global_tool,
+        schedule(0, _(outer(block)) -> set(pos(block), block));
+        _draw_surf(1);
+        global_points=[];
+        global_points+=pos(block);
+        global_sets:(length(global_sets))=global_points;
+        _normalise_set();
+        _draw_surf(10000)
```
Firigion, line 48: I understand the idea of just drawing for a long time, but the proper way to do this is to call a function that will then schedule itself. Something like 
```
_draw_surf_tick() -> (
  _draw_surf(12);
  if(global_draw_surf, schedule(10, '_draw_surf_tick') )
);
```
where here 10 is how often the surf is updated and 12 is a number just a bit over 10 so you don't get flickering in slower mashines. Both numbers acna of course be global values so thye are easily set.

programs/loft.sc
```diff
@@ -0,0 +1,356 @@
...
+		},
+        'replace'->{'type'->'term','suggest'->['air','water','lava']}
+    }
+};
+
+global_tool='phantom_membrane';
+global_block='diamond_block';
+global_replace=null;
+global_sets=[]; //array containing current information about the m splines created
+global_points=[]; //vector of maximum n points of the current spline that is being created by the player
```
Firigion, line 26: No need to set this here if you are gonna set it again inside `__on_player_breaks_block`. But maybe it's good to have it here so someone reading the code knows about all the global variable syou are gonna use.

programs/loft.sc
```diff
@@ -0,0 +1,356 @@
...
+            print('operation cancelled');
+            global_sets=[];
+            global_cset=[];
+            global_points=[]
+        )
+    )
+);
+
+__on_player_right_clicks_block(player, item_tuple, hand, block, face, hitvec)->( //expands the spline adding new points
+    if((get(item_tuple,0)==global_tool)&&(hand=='mainhand')&&(global_points!=[]),
```
Firigion, line 62: in general, get calls can be replaced by the get operator `:`. So `get(item_tuple,0)` is absolutely equivalent to `item_tuple:0`. Just FYI.

programs/loft.sc
```diff
@@ -0,0 +1,356 @@
...
+        _draw_surf(1);
```
Firigion, line 63: if this call here here is just so that you cancel the old shape and then can call a new one, this would be immediately solved by the suggestion above regarding schedule.

programs/loft.sc
```diff
@@ -0,0 +1,356 @@
...
+        global_points=[];
+        global_points+=pos(block);
+        global_sets:(length(global_sets))=global_points;
+        _normalise_set();
+        _draw_surf(10000)
+    );
+    if(item_tuple!=global_tool,
+        if(global_sets!=[],
+            _draw_surf(1);
```
Firigion, line 52: in the suggestion above, this would be replaced by setting global_draw to false.

programs/loft.sc
```diff
@@ -0,0 +1,356 @@
...
+        )
+    )
+);
+
+_draw_spline(points,number_of_tick)->( //a function used to render a spline given it's vector of points
+    if(length(points)==1, //in case the spline has only one point (not a spline obviously, just a point)
+        p1=points:0;
+        draw_shape('sphere',number_of_tick,{'center'->p1+0.5,'radius'->1})
+    );
+    if(length(points)==2, //in case the spline has only 2 points (a rect line)
```
Firigion, line 211: *straight line

programs/loft.sc
```diff
@@ -0,0 +1,356 @@
...
+    if(lvmax>0, dv=0.7/lvmax, dv=1);
+    if(lumax>0, du=0.7/lumax, du=1);
+    while(v<=1,1000,
+        u=0;
+        while(u<=1,1000,
+            point=(p00*(1-v)^2+2*a0*v*(1-v)+p01*v^2)*(1-u)^2+     //point of the surface s
+                (2*b0*(1-v)^2+4*c*v*(1-v)+2*b1*v^2)*u*(1-u)+
+                (p10*(1-v)^2+2*a1*v*(1-v)+p11*v^2)*u^2;
+            p=[round(get(point,0)),round(get(point,1)),round(get(point,2))];
+            schedule(0,'_set_block',p);
```
Firigion, line 297: why the schedule here?

programs/loft.sc
```diff
@@ -0,0 +1,356 @@
...
+);
+
+paste(Block,replace)->(
+    global_block=Block;
+    global_replace=replace;
+    m=length(global_cset);
+    n=length(global_cset:0);
+    if((n>1)&&(m>1),
+        c_for(i=0,i<n-1,i=i+1,
+            game_tick(10);
```
Firigion, line 317: what was your intention with this line?

programs/loft.sc
```diff
@@ -0,0 +1,356 @@
...
+                    global_c:j:i
+                )
+            )
+        );
+        schedule(1,'_add_to_history');
+        _draw_surf(1);
+        global_sets=[];
+        global_cset=[];
+        global_points=[]
+    )
```
Firigion, line 337: Some sort of feedback would be welcome here. Something like "such and such pasted" or something like that.

Firigion (changes requested):
I'm not totally sure how to use the app. I read through the code (most of it, i skipped the functions tha tdo the math because i didn't feel like reading them and it looks like they work) and figured out that grabbing a phantom membrane and left clicking a block would start a new draw function and right clicking other blocks would create the spline (line)

image

Thing is, running `/loft paste` didn't palce the blocks, I'm not sure as to why. The other thing I'm not sure how to do is make a surface. All I could do was a line.

All in all, from the shocases you did, the app looks good, but I think it needs a bit of work on the UX side [for example, why not use the `__on_player_swigs_hand` and `__on_player_uses_item` events so that you can do stuff in mid air isntead of needig to pre-place blocks (since placing a block while creating a spline will cancel it)] and it absolutely needs a help function that describes at least how to use it.

That baing said, an absolutely astonishing first app!

programs/loft.sc
Firigion: Oh, also: I just noticed you dropped this in the `programs` folder instead of `programs/building`. Please move it there and add a line to the readme in that folder (ask for help if you don't know how to).

programs/loft.sc
Cfer98: Ok, I'll try to moove the pull request in the right section.
About the surface selection, you have to create 2 or more splines to have a proper surface. As said in the script, left click starts a new spline and right click expands the last spline created. Then after you have selected at least 2 splines, the paste command should be successfull. The script basically just do a loft using those splines created by the player and that generates a curve surface. If there is only one spline the paste will fail. It will olso fail if you create more splines but each of them is composed by only one point. There is a way to paste a single spline and that is to select 2 splines using the same control points. So the first spline is exactly equal to the second one.
Hope this explenation will help

programs/loft.sc
Cfer98: So, can global variables be declared in one function of the script?

programs/loft.sc
Cfer98: Oh, I didn't know that

programs/loft.sc
Cfer98: Because i didnt want to cause a lagspike when setting blocks. I though it was a good thing

programs/loft.sc
Cfer98: Oh, ops. I forgot to delete line 317. I added the delay when I was trying to debug the script

PR #303, opened 2022-03-19
Add loft app to the building section

Cfer98:
A tool to make splines and merge them in a one surface with a loft function. Use the phantom membrane as a wand and select blocks to define the surface points. Start a new spline by left clicking on block and extend it by right clicking on blocks. Create in this way at least 2 splines to have a proper surface that can be pasted with `/loft paste`.


Cfer98 (commented):
I accidentally added and removed the same thing inside the README.md so nothing changed here (I think)

PR #305, opened 2022-03-19
Balanced Waystones 

opsaaaaa:
This app makes lodestones into waystones players can teleport to after visiting.
The goal is to add waystones to the game in a balanced and natural feeling way.

I often find teleports/ warps on smp servers feel overpowered and so im am experimenting with a middle ground. 

When a player clicks a lodestone it displays a list of waystones the player has perviously visited.
The icon for each location is the block the lodestone was placed on.
Waystone can be named by naming the lodestone in an anvil or with a name tag after the fact.

Configurable Features:
- Teleporting to a waystone costs an ender_eye.
- A players waypoints get reset on death.
- Activating a waystone requires 6 copper blocks in a 5x5 area around the lodestone.
- by default waystones only work in the overworld dimension.
- by default waystones can't cross dimensions.

A few things still feel a clunky but I'm running it on a server and I think its working well enough.

(also yes this is the same world, my testing world started generating normal chunks)


programs/survival/waystones.sc
```diff
@@ -0,0 +1,310 @@
...
+        ''->'open_waystones_screen',
+        'setting offering <item>'->['setting_command','offering'],
+        'setting offering'->['setting_read','offering'],
+        'setting dimensionalCrossing <bool>'->['setting_command','dimensional_crossing'],
+        'setting dimensionalCrossing'->['setting_read','dimensional_crossing'],
+        'setting clearWaypointsOnDeath <bool>'->['setting_command','clear_waypoints_on_death'],
+        'setting clearWaypointsOnDeath'->['setting_read','clear_waypoints_on_death'],
+        'setting enabledDimensions <dimensions>'->['setting_command','enabled_dimensions'],
+        'setting enabledDimensions'->['setting_read','enabled_dimensions'],
+        'setting structureMaterial <dimensions>'->['setting_command','structure_material'],
```
altrisi, line 43: Missed this

programs/survival/waystones.sc
```diff
@@ -0,0 +1,310 @@
...
+    display_title(p, 'actionbar', format('wb '+m));
+);
+
+setting_read(key) -> (
+    print(player(), str('%s: %s', key, global_settings:key));
+);
+
+setting_command(val, key) -> (
+    global_settings:key = val;
+    print(player(), str('%s: %s', key, global_settings:key));
```
altrisi, line 85: Maybe "%s was set to %s"? Else this has the same output as read.

programs/survival/waystones.sc
```diff
@@ -0,0 +1,310 @@
...
+                    uuid = p~'uuid';
+                    _read_player_waypoints(uuid);
+                    _mark_player_waypoint(pos, p~'pos', uuid); 
+                    sound( 'minecraft:block.amethyst_block.hit', block );
+                    _open_waypoint_screen(p, pos, uuid);
+                    _write_player_waypoints(uuid);
+                );
+            ,
+                _app_message(p, str('Place the lodestone near a %d block %s structure to activate',
+                    global_settings:'structure_size',global_settings:'structure_material'));
```
altrisi, line 138: It can get a little confusing when you place the lodestone first, then the structure until you know it.

programs/survival/waystones.sc
```diff
@@ -0,0 +1,310 @@
...
+        in_dimension(dimension, 
+            run(str('tp %s %d %d %d', p~'command_name', ...(pos + [0.5, 0, 0.5]) ));
+        );
+        schedule( 2, _(p, pos, dimension)->(
+            in_dimension(dimension, 
+                sound( 'item.trident.thunder', pos );
+                particle('totem_of_undying', p~'pos'+[0,1,0], 100, 1, 0.1);
+            );
+        ), p, pos, dimension);
+        _app_message(p, global_waystones:point:'name');
```
altrisi, line 180: Maybe something like "Teleported to %s" would be clearer, just the name looks a bit weird.

programs/survival/waystones.sc
```diff
@@ -0,0 +1,310 @@
...
+    // merge waypoints and waystones data into icons.
+    for(pairs(global_waypoints:uuid),
+        if( global_waystones:(_:0), 
+            icons += ([_:0,global_waystones:(_:0),_:1]);
+        , 
+            bad_keys += _:0;
+        );
+    );
+
+    // Clear player waypoints with missing waystone entires
```
altrisi, line 234:
```suggestion
    // Clear player waypoints with missing waystone entries
```

Also talking about clearing entries, I've noticed that waypoints destroyed by non-direct player actions (such as explosions) are kept _and_ working. IMO they should be removed, maybe not every time someone opens the menu if it eats performance given it has to load the area, but it could be checked before tp at least.

programs/survival/waystones.sc
```diff
@@ -0,0 +1,310 @@
...
+_print_icons_to_screen(screen, icons) -> (
+    for(slice(sort_key(icons, _:1:'name'),0, min(length(icons),53)),
+        stone = _:1;
+        nbt = nbt({
+            'pos'->_:2,
+            'point'->_:0,
+            'dimension'->stone:'dimension' || 'overworld',
+            'display'->{
+                'Lore'->[escape_nbt(str('"%d, %d, %d"', _:0))],
+                'Name'->escape_nbt( str('"%s"', stone:'name'))
```
altrisi, line 271: I think it'd be cool to have some formatting applied to the raw data. Right now them getting italic or the amazing vanilla purple lore isn't that great IMO, and this is basically the most visible part of the app.

programs/survival/waystones.sc
```diff
@@ -0,0 +1,310 @@
...
+            }
+        });
+        try(
+            if(stone:'icon'=='air', throw('Dont set inventory slot to air'));
+            inventory_set(screen, _i, 1, stone:'icon', nbt);
+        , 'exception', 
+            inventory_set(screen, _i, 1, 'lodestone', nbt);
+        );
```
altrisi, line 279: Can't this be replaced with an `if`, without an exception in the middle? Or is there some other thing it's catching there?

programs/survival/waystones.sc
```diff
@@ -0,0 +1,310 @@
...
+        _app_message(p, str('The Waystone requires an %s offering', global_settings:'offering'));
+        sound( 'item.trident.return', p~'pos' );
+        false;
+    );
+);
+
+_warp_player(p, pos, dimension, point)->(
+    if(_has_required_item(p),
+        in_dimension(dimension, 
+            run(str('tp %s %d %d %d', p~'command_name', ...(pos + [0.5, 0, 0.5]) ));
```
altrisi, line 172: Small thing: The player will suffocate if the waystone spawn has been covered! Is it worth it to have logic to find a safe place?

programs/survival/waystones.sc
```diff
@@ -0,0 +1,310 @@
...
+
+__on_player_places_block(p, item, hand, block)->(
+    if(item:0=='lodestone',
+        pos = pos(block);
+        floor = str(block(pos:0, pos:1 - 1, pos:2));
+        if(_has_required_structure(pos),
+            global_waystones:pos = {
+                'icon'->floor,
+                'dimension'->current_dimension(),
+                'name'->_get_name_from_nbt(item:2) || floor+' Waypoint'
```
altrisi, line 96: Would look a lot better if you could get the block name, but I don't think scarpet can do that atm? Possible suggestion to scarpet tbh.

altrisi (approved):
Looks amazing! Loving it so far!

Left some small suggestions, nothing really major. Feel free to apply the changes, but lgtm as is!

programs/survival/waystones.sc
opsaaaaa: Woops

programs/survival/waystones.sc
opsaaaaa: Yeah, the script really should allow you to add the structure after the fact. perhaps it could check if you added a structure when you click in the stone. Ill look into that.

programs/survival/waystones.sc
opsaaaaa: Thats a very good point. I had not even thought about explosions. Ill add a check before the player teleports.

programs/survival/waystones.sc
opsaaaaa: the Icon is a block, and inventory_set expects an item. they don't always match up 100% of the time. when inventory set fails it throws an error. for example if you place the stone above lava, it would complain about item not found. when you pass it air tho it does not throw an error, but it fails down the line when you try to render air with lore and a name. 

I looked at ways to do this without a try fail block, and this seemed to be the easiest. It would be nice if blocks like lava mapped to a lava bucket but i don't really want to implement that. It would also be nice if you had a way to use items like a bone for the icon, but i have not found a nice way to add that feature. If i come back to this with improvements that will be one of them.

programs/survival/waystones.sc
opsaaaaa: Probably not in my opinion. My current solution saves the last location they opened the way stone from. That should be safe, aside from another player filling in the area with blocks. If this was used on a public server with poorly behaved players that could be an issue. It could also be a amusing prank, similar to filling in the area a player logged out from, or boxing in a afk player.

Finding a safe location would probably involve either randomly picking block around the stone or looping over the area around the stone. I think that's a lot of code for something that's not going to happen very often.

programs/survival/waystones.sc
opsaaaaa: I am trying this, and as far as i can tell formatting the item name or lore with the `format` scarpt method is not working. It looks possible in default mc but it seems slightly more complicated then I want to think about rn.

programs/survival/waystones.sc
opsaaaaa: i changed it to term instead of block to allow for less strict inputs like `plank`, `copper`, `log`, `birch` instead of strict block names.

programs/survival/waystones.sc
opsaaaaa: it now attempts to activate a lodestones instead of rendering that message.

programs/survival/waystones.sc
opsaaaaa: I am putting this feature on hold for that fix for the screen callback position bug.

PR #306, opened 2022-04-03
Adding Classes to scarpet.

Ghoulboy78:
This is a continuation from [this discord conversation](https://discord.com/channels/882822986795716608/953977261827170355), and the code I wrote there.

In this PR, I also rename `heap.sc` to `min_heap.sc`, and add `max_heap.scl` as a showcase of what could be done with classes.


BisUmTo (approved):
I spent 1 hour learning how should max_heap actually work 🤣

The classes.scl is really great, I'll surely use it for my projects

Ghoulboy78:
Because otherwise you end up just having 1 map and so one object, instead of multiple instances of the same class of object. Taking a copy means that the original is unchanged and the `new_object` method returns a separate map to the one we input.
I came across this problem when rewriting matrix as a class to make it easier, and then realised that I was having this issue.

Ghoulboy78:
BTW Im working on inheritance rn. It's not too hard and I've got it mostly working already, just need to figure out whether inheriting a class which already inherits from another class works or not.
Definitely parametric classes (At least I think that's what they're called? I'm talking about things like `class MyClass<A>` or something like that)

Ghoulboy78:
Ok so I've finished inheritance. I will be adding extensive documentation to Readme.md cos currently it's got like 1 line tbh.

Ghoulboy78:
I'm going to make all classes have a couple fundamental methods, which I will put into an `'object'` class

programs/fundamentals/classes.scl
```diff
@@ -0,0 +1,140 @@
...
+global_interface_class = new_class('Interface', {
+    '__init__'->_(self)->throw('Cannot initialize an interface class'),
+});
+
+//Some template classes. These may in time be moved to a separate file
+global_iterator_class = new_class('Iterator', {
+    'has_next'->_(self)->null,
+    'next'->_(self)->null,
+    'for_each'->_(self, function)->(
+        while(call_function(self, 'has_next'), 0x7FFFFFFFFFFFFFFF,
```
ch-yx, line 132: use c_for to remove wierd 0x7FFFFFFFFFFFFFFF?

opsaaaaa (approved):
Looking at the code. I think this is good for what it is!
It looks thought out. I'm not seeing any obvious better approaches in the few minutes i've thought about it.

using `call()` for everything feels clunky, but thats how you have to do it.

I think large scripts are the once's that would benefit from this. 
I remember bemoaning not having classes while writing the spellbook script.
Ill try this out more thoroughly if/when I rewrite it. (or any other larger scripts)

Ghoulboy78:
Still gotta figure out /script download stuff

Ghoulboy78:
Yeah, I don't think the /script download stuff is so important, I'm just gonna put this up for review as it is, cos other stuff depends on it.

PR #307, opened 2022-04-04
Adding block_broken event

### LINKED ISSUE #298: Add compatibility between blocl-breaking apps
**Is your feature request related to a problem? Please describe.**
I would like the blocks that I break with hammer.sc to be picked up with auto_pickup.sc and same with blocks broken by veinminer.sc

**Describe the solution you'd like**
Basically I would like all those apps to trigger and listen for an event called `block_broken` or something, which has the exact same parameters as `__on_player_breaks_block`. The reason being that if apps trigger `__on_player_breaks_block`, other apps may trigger involuntarily, so `block_broken` event circumvents that by making sure that apps which listen for it know the implications of it being fired and treat it specially.

**Describe alternatives you've considered**
My first solution was to just trigger `__on_player_breaks_block`, until I realised that it was a bad idea.

**Additional context**
I may implement this myself, but it will take a little while, hence why I'm making this issue.

Ghoulboy78:
This will allow block breaking apps to signal to other apps that blocks have been broken, as well as allow apps to receive this information. Fixes #298 
Todo;
- [ ] Readme.md update
- [ ] Somehow modify scoreboards so auto_pickup.sc works with vacuum.sc

### DIFF
--- a/programs/survival/auto_pickup.sc
+++ b/programs/survival/auto_pickup.sc
@@ -1,17 +1,14 @@
 // stay loaded
-__config() -> (
-   m(
-      l('stay_loaded','true')
-   )
-);
+__config() -> {
+	'stay_loaded'->'true'
+};
 
 // seems like drops are generated later, before the callback is made
 // so we execute moving items at the end of the tick
 __on_player_breaks_block(player, block) -> 
 (
-	schedule(0,'_move_items_to_inventory', player, pos(block))
+	signal_event('block_broken', player, pos(block));
 );
-
 
 // (_ ~ 'pickup_delay') == 10 check is optional
 // it will only pick up items that are 'pick-broken'.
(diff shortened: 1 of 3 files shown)


ch-yx:
i m thinking, should there be a flag about what the game is currently doing. 
for example, it can tell me the item entity s spawning is because of a players mining or schedule ticks (such as cactus).

GaelCLuna:
auto_pickup and instamine_deepslate are incompatible. Instamine_deepslate works but the items don't get put into ur inventory

altrisi:
Is this ready for review? Wondering given there's stuff in the TODO but the PR isn't a draft.

I'll mark it as a draft because of that, but unmark if it's ready.

Ghoulboy78:
yeah, this is still WIP. I've just not had time do tackle it.

PR #308, opened 2022-04-07
Add Dynamic Render Distance Display app to utilities

SulpherStaer:
image

An app that displays the current server render distance in the footer of the player list.
This way the players always know the 'current' render distance. (some time delay)

This is intended to work together with a dynamic render distance mod, like the one made by [Henkelmax](https://github.com/henkelmax/dynamic-render-distance).

The script has commands to configure the min and max render distance. The min/max is used in the color coded calculations, which use the same colors & calculation that carpet does for tps. The colors however are inverted. Where with tps 0 is green, and 50 is magenta, the render distance display has min as magenta, and max as green.

The amount of time it takes before the script checks the render distance again is customizable.

The script also saves the configuration on exit, to use it when the script is loaded again. Commands are locked behind op permissions.

A list of all commands with this script:

- /renderdistance display maximumRenderDistance \<max\> - _sets the maximum value._
- /renderdistance display minimumRenderDistance \<min\> - _sets the minimum value._
- /renderdistance display limit \<min\> \<max\> - _allows setting the minimum and maximum with a single command._
- /renderdistance refreshTimer \<seconds\> - _sets the time delay between refreshes of the display._
- /renderdistance display update - _Instantly updates the display for every player online._
- /renderdistance display settings - _shows the current min & max, and refresh time delay._

nb. the reason for the command being `renderdistance display` and not just `renderdistance` is because the mod from henkelmax also uses `renderdistance` as its root command.

### DIFF
--- a/programs/utilities/renderdistance.sc
+++ b/programs/utilities/renderdistance.sc
@@ -10,6 +10,7 @@ global_server_min_render_distance = 10;
 
 __config() -> {
     'scope' -> 'global',
+    'allow_command_conflicts' -> true,
     'command_permission' -> 'ops',
     'commands' -> {
         'display maximumRenderDistance <max>' -> 'set_display_max_render_distance',
--- a/programs/utilities/renderdistance.sc
+++ b/programs/utilities/renderdistance.sc
@@ -1,0 +1,104 @@
+// Displays the current render distance for servers with a dynamic render distance mod. (Like the one made by Henkelmax)
+// When a player joins they can see the render distance at the bottom of the player list.
+// It refreshes every 1200 ticks (1 minute) and color codes the distance according to carpet TPS logging colors.
+// The refresh time of the display and the min/max render distance can be altered with commands. This will be saved.
+// Made by Sulpherstær with aid from people in the carpet discord. (notably: Surf, altrisi, rv3r, Firigion, and Crec0)
+
+global_refresh_time_in_ticks = 1200;
+global_server_max_render_distance = 32;
+global_server_min_render_distance = 10;
+
+__config() -> {
+    'scope' -> 'global',
+    'commands' -> {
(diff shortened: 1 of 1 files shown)


programs/utilities/renderdistance.sc
```diff
@@ -0,0 +1,105 @@
...
+// When a player joins they can see the render distance at the bottom of the player list.
+// It refreshes every 1200 ticks (1 minute) and color codes the distance according to carpet TPS logging colors.
+// The refresh time of the display and the min/max render distance can be altered with commands. This will be saved.
+// Made by Sulpherstær with aid from people in the carpet discord. (notably: Surf, altrisi, rv3r, Firigion, and Crec0)
+
+global_refresh_time_in_ticks = 1200;
+global_server_max_render_distance = 32;
+global_server_min_render_distance = 10;
+
+__config() -> {
```
Ghoulboy78, line 11: If you want to be able to use this script in conjunction with another mod which adds the `/renderdistance` command, you should add `'allow_command_conflicts'->true` to the config or it may not work

SulpherStaer: At the moment there does not appear to be a conflict at all. But I'll add the line still. 👍

Ghoulboy78 (approved):
Ok, looks good to me

SulpherStaer:
There appears to be a bug with either this script, or with carpet itself, but the config rule `'command_permission' -> 'ops',` does __not__ work at all. Tried setting it to 2 as well. Did /reload's, did /script unload/loads - did server restarts. It does not limit command permission at all.

altrisi:
Note that adding to existing commands is not currently supported, it working is a bug IINW, it shouldn't let you load the app (the check is currently only in the legacy path, was forgotten for the new one). Should be allowed? Good question, has pros and cons.

PR #312, opened 2022-04-25
Removing return statements from gm.sc

Ghoulboy78:


BisUmTo (approved):

opsaaaaa (approved):

altrisi:
Note that those make no performance difference given they're at the end of the function.

Ghoulboy78:
ah. nvm then

PR #313, opened 2022-04-25
Properly setting clear weather in sleep_ignore_fake_players.sc

### LINKED ISSUE #241: sleep_ignore_fake_players.sc make that a thunderstorm appends next day
When loaded every time you sleep a thunderstorm will append next day. Sleeping to end the storm result in another storm next day and so on.

1. Load the script in a game
2. Wait and sleep when you can
3. Wait for around mid day and a thunderstorm should start

**Expected behavior**
When sleeping with fake players around there should be a normal chance of having bad weather.

**Environment**
 - Carpet version: Carpet Jar Mod v19_06_04 with Scarpet v1.3 on Fabric 0.7.4
 - Minecraft version: Minecraft 1.17.1
 - Other mods: malib, litematica, item scroller, tweakaroo, modmenu

Ghoulboy78:
Fixes #241 
Using `weather('clear', round(rand(0x7FFFFFFF))` instead of `run('weather clear')` which always does 5 minutes (or 6000 ticks)

### DIFF
--- a/programs/survival/sleep_ignore_fake_players.sc
+++ b/programs/survival/sleep_ignore_fake_players.sc
@@ -17,7 +17,7 @@ in_dimension('overworld',
             global_allPlayerSleep = false;
             l = day_time() + 24000;
             day_time(l - l % 24000);
-            run('weather clear');
+            weather('clear', round(rand(0xFFFFFFFF))
         );
     );
 
--- a/programs/survival/sleep_ignore_fake_players.sc
+++ b/programs/survival/sleep_ignore_fake_players.sc
@@ -17,7 +17,7 @@ in_dimension('overworld',
             global_allPlayerSleep = false;
             l = day_time() + 24000;
             day_time(l - l % 24000);
-            if(weather('rain')>0, weather('clear', round(rand(0xFFFFFFFF)))
+            if(weather('rain')>0, weather('clear', round(rand(0x7FFFFFFF)))
         );
     );
 
--- a/programs/survival/sleep_ignore_fake_players.sc
+++ b/programs/survival/sleep_ignore_fake_players.sc
@@ -17,7 +17,8 @@ in_dimension('overworld',
             global_allPlayerSleep = false;
             l = day_time() + 24000;
             day_time(l - l % 24000);
-            if(weather('rain')>0, weather('clear', round(rand(0x7FFFFFFF))))
+            // doing this also sets rain and thunder to 0 because of scarpet impl, exactly what vanilla does when sleeping (from ServerWorld in yarn)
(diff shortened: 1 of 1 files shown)


BisUmTo:
According to the wiki:
```
21w44a: Sleeping now only resets the weather cycle if it is currently raining or snowing.
```

Ghoulboy78:
k cool, I'll modify it.

Ghoulboy78:
Also changed to using 0x7FFFFFFF not 0xFFFFFFFF cos 0xFFFFFFFF as an integer is -1

programs/survival/sleep_ignore_fake_players.sc
```diff
@@ -17,7 +17,7 @@ in_dimension('overworld',
             global_allPlayerSleep = false;
             l = day_time() + 24000;
             day_time(l - l % 24000);
-            run('weather clear');
+            if(weather('rain')>0, weather('clear', round(rand(0x7FFFFFFF)))
```
opsaaaaa, line 20:
```suggestion
            if(weather('rain')>0, weather('clear', round(rand(0x7FFFFFFF))))
```
missed a `)`

opsaaaaa (changes requested):
After adding the ) it works as intended.

altrisi (approved):
Makes sense

rv3r:
Weather resetting works as intended, but I won't approve since I don't know how long it should be reset for. If anyone can show that the random number range the code uses if correct, I'll approve this.

programs/survival/sleep_ignore_fake_players.sc
```diff
@@ -17,7 +17,7 @@ in_dimension('overworld',
             global_allPlayerSleep = false;
             l = day_time() + 24000;
             day_time(l - l % 24000);
-            run('weather clear');
+            if(weather('rain')>0, weather('clear', round(rand(0x7FFFFFFF))))
```
altrisi, line 20:
```suggestion
            // doing this also sets rain and thunder to 0 because of scarpet impl, exactly what vanilla does when sleeping (ServerWorld in yarn)
            if (weather('rain') > 0, weather('clear', weather('clear')))
```

altrisi (changes requested):

programs/survival/sleep_ignore_fake_players.sc
rv3r: Yep running `weather('clear')` just before this line returns `0`.

altrisi (approved):

rv3r (approved):
Change to weather reset makes sense, so I'll approve

PR #316, opened 2022-05-09
Stop player from exiting World Border using /cam

LarveyOfficial:
Changes made by @altrisi, this prevents someone from leaving a world's world border using /cam

### DIFF
--- a/programs/survival/cam.sc
+++ b/programs/survival/cam.sc
@@ -30,6 +30,7 @@ __command() ->
    p = player();
    current_gamemode = p~'gamemode';
    if ( current_gamemode == 'spectator',
+      entity_event(p, 'on_move', null);
       if (config = __get_player_stored_takeoff_params(p~'name'),
          __remove_camera_effects(p);
          __restore_player_params(p, config);
@@ -57,6 +58,17 @@ __command() ->
             __store_player_takeoff_params(p);
             __turn_to_camera_mode(p);
             display_title(p, 'actionbar', format('y Entered camera mode'));
+            entity_event(p, 'on_move', _(p, vel, p1, pos) -> (
+    			   center = system_info('world_center');
+   				size = system_info('world_size');
+    			   positive = center + size;
+    			   negative = center - size;
+    			   pos = p ~ 'pos';
+    			   if (pos:0 > positive:0 || pos:0 < negative:0
+      			|| pos:2 > positive:2 || pos:2 < negative:2,
+                  __command();
+    			   );
+			   ));
          )
       ));
    );
--- a/programs/survival/cam.sc
+++ b/programs/survival/cam.sc
@@ -19,6 +19,8 @@ __assert_player_can_cam_out(player) ->
 // none of your business below
 
 
+global_exit_messages = ['y Exited camera mode', 'r You tried to '];
+global_exit_message = 0;
 
 
 __config() -> {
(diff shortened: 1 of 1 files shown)


Ghoulboy78:
Well currently it does this by turning off the cam. Would it be better to just block them from going further? Also if ur going to do this, then perhaps add a note in the Readme.md as well as an action bar message telling the player like "you went out of the world border" or smth?

LarveyOfficial:
Well currently it doesn't do anything. It just allows the spectator to infinitely go past the border. This just instantly evokes /cam to reset their location and gamemode

Ghoulboy78:
I mean that in the current version of this app it just shuts off cam when the player tries to exit the world border. I'm proposing just blocking them at the world border, and also giving them a message saying that "you're trying to exit the world border" or smth

LarveyOfficial:
We considered that, but it can be messy at times. After some discussion we came to the conclusion that just disabling cam at the border and having them tp back to their original position would overall be better.

LarveyOfficial:
> We considered that, but it can be messy at times. After some discussion we came to the conclusion that just disabling cam at the border and having them tp back to their original position would overall be better.

This also prevents annoying players from trying to "bug" past the border even if they can't, although a message stating they can't pass the border may be a good idea.

Ghoulboy78:
thst seems like a better idea

LarveyOfficial:
Which seems like a better idea?

Ghoulboy78:
Turning off cam and also sending a message to the player to tell them why

LarveyOfficial:
Ah yes that's probably the best idea. But like I pointed out at the start of the PR I don't know the syntax, the changes where made by @altrisi

programs/survival/cam.sc
```diff
@@ -57,6 +58,17 @@ __command() ->
...
+            entity_event(p, 'on_move', _(p, vel, p1, pos) -> (
+    			   center = system_info('world_center');
+   				size = system_info('world_size');
+    			   positive = center + size;
+    			   negative = center - size;
+    			   pos = p ~ 'pos';
+    			   if (pos:0 > positive:0 || pos:0 < negative:0
+      			|| pos:2 > positive:2 || pos:2 < negative:2,
+                  __command();
+    			   );
```
Ghoulboy78, line 70:
```suggestion
    			   if (pos:0 > positive:0 || pos:0 < negative:0
      			|| pos:2 > positive:2 || pos:2 < negative:2,
                  __command();
                  display_title(p, 'actionbar', format('r You tried to exit the world border'));
    			   );
```

This change ought to work.

Ghoulboy78: oh wait no, it'll give the double message

Ghoulboy78:
I have made the changes

Ghoulboy78 (approved):
I can say it works as intended, tho my code looks kinda ugly

programs/survival/cam.sc
```diff
@@ -19,6 +19,8 @@ __assert_player_can_cam_out(player) ->
 // none of your business below
 
 
+global_exit_messages = ['y Exited camera mode', 'r You tried to '];
```
altrisi, line 22: You tried to... tried to what?
```suggestion
global_exit_messages = ['y Exited camera mode', 'r You tried to exit the world border'];
```

LarveyOfficial:
Looks good!

altrisi:
I'm not going to give an approval to this btw, it would feel like cheating given I wrote most of it (and authors can't self approve).

So this is missing one approval.

LarveyOfficial:
> I'm not going to give an approval to this btw, it would feel like cheating given I wrote most of it (and authors can't self approve).
> 
> So this is missing one approval.

That's fair, we can probably wait for another dev to approve it. I understand your reasoning.

PR #182, opened 2021-02-13
Fix Signs for water streams

Ghoulboy78:
Fixes #140

PR #320, opened 2022-05-22
Added my changes to correct signs for water streams

### LINKED ISSUE #184: #182 removed some commits from the Signs for water streams app
When making and reviewing #182, no one noticed the fix was based on a previous version of the app, therefore letting commits a07f7f391804a033da60bde4b6d402af07fda55e and f1da9ea3961566b3318e0925f15972e75a143f6e outside of the current app.

Most notably, that removed the heuristics that made the app work with modded signs. It also removed a dupe fix when placing signs from offhand and a QOL for placing in creative without using items (not really tested those last ones though).

Why I think it was based on the old version (probably 3e0b0ebe5c0d93cfc833add96789902d40adc4be):

Scarpet version was dropped down, when it went up in those commits: https://github.com/gnembon/scarpet/commit/38ff8c574eb0ee7abb87b2bff7410bdbdb32c284#diff-8bb66dc9e8f062b283436ed4626c5d88f606d4ce010de8c2c3a7fdf09d8d59f7R6-R8

Also the line specifying the update is also gone, plus some of the old formatting is back.

Pointed out by the author of those commits in https://github.com/gnembon/scarpet/pull/182#issuecomment-780252775.

Ghoulboy78:
Fixes #184 
As altrisi mentioned, in #182 I managed to take out two commits, which properly depleted the player's hand stock.

### DIFF
--- a/programs/survival/signs_for_water_streams.sc
+++ b/programs/survival/signs_for_water_streams.sc
@@ -24,7 +24,8 @@ global_item_to_block = {
   'oak_sign'-> 'oak_wall_sign',
   'spruce_sign'-> 'spruce_wall_sign',
   'crimson_sign'-> 'crimson_wall_sign',
-  'warped_sign'-> 'warped_wall_sign'
+  'warped_sign'-> 'warped_wall_sign',
+  'mangrove_sign'-> 'mangrove_wall_sign'
 };
 
 __on_player_right_clicks_block(player, item_tuple, hand, block, face, hitvec) ->(
--- a/programs/survival/signs_for_water_streams.sc
+++ b/programs/survival/signs_for_water_streams.sc
@@ -43,7 +43,7 @@ __on_player_right_clicks_block(player, item_tuple, hand, block, face, hitvec) ->
       clicked_block_pos = map(pos(block), str('%.2f',_));
       offset_by_one = pos_offset(block,face,1); // Without the offset, the block you clicked on will be turned into a sign.
       // Make sure there's a replaceable block there, otherwise you could auto-destroy blocks with this.
-      replaceable_materials = m ('air', 'water', 'lava', 'grass', 'sea_grass');
+      replaceable_materials = m('air', 'water', 'lava', 'vegetation', 'sea_grass');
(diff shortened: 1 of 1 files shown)


opsaaaaa (approved):
Works when i tested it.
And i didn't see any obvious issues with the code.

programs/survival/signs_for_water_streams.sc
```diff
@@ -1,52 +1,57 @@
...
 global_item_to_block = {
-	'acacia_sign'-> 'acacia_wall_sign',
-	'birch_sign'-> 'birch_wall_sign',
-	'dark_oak_sign'-> 'dark_oak_wall_sign',
-	'jungle_sign'-> 'jungle_wall_sign',
-	'oak_sign'-> 'oak_wall_sign',
-	'spruce_sign'-> 'spruce_wall_sign',
-	'crimson_sign'-> 'crimson_wall_sign',
-	'warped_sign'-> 'warped_wall_sign'
+  'acacia_sign'-> 'acacia_wall_sign',
```
MeeniMc, line 20: Looking good, you forgot to add the new 1.19 sign to this list though

MeeniMc (approved):

programs/survival/signs_for_water_streams.sc
Ghoulboy78: I will, tho if you look this map isn't actually used anywhere (it's one usage is commented out). I don't know why, cos I didn't write that code.

altrisi (approved):
Works great!

altrisi:
Easy review if someone has a moment.

rv3r (changes requested):
Using `material()` and `grass` leads to a sign replacing a grass block. Replicate using the screenshot below by placing the sign against the side of the wall nearest the grass block. 

Additionally, the block known as just `'grass'` has material `vegetation` if the author's intent was for signs to replace grass.
Since an ender chest, apparently, has no `inventory_size()`, sign editing GUI is skipped as though is were a regular block. If this is the author's intent, I am okay with that.
Other than that, works well!

rv3r (approved):
Works as described, no bugs discovered

PR #321, opened 2022-05-22
Making auto_lighter finally light the ground

### LINKED ISSUE #317: auto_lighter.sc not does not light the surface by default
**Describe the bug**
I was trying the `auto_lighter.sc` app and then i just afk like 10 minutes, after that i noticed that the app wouldnt light the surface (i was on creative), downgrading the app and using this guide https://github.com/gnembon/fabric-carpet/wiki/How-to-spawn-proof-with-auto_light.sc should work

**To Reproduce**
1. Go to a world and download the latest version of the app
2. Load `auto_lighter`
3. Use the torch and start the app
4. Wait a while and you'll see how it never lights the surface only caves

**Expected behavior**
It should work as the previous version with the guide i mencionated earlier but with the conditional that specificies in the new version if it should ligh the surface or not

**Environment**
 - Carpet version: 1.4.69+v220331
 - Minecraft version: 1.18.2

**Additional context**
It was a minecraft server, and it happens on singleplayer too

Ghoulboy78:
Fixes #317 
This time I actually tested it thoroughly.

### DIFF
--- a/programs/survival/auto_lighter.sc
+++ b/programs/survival/auto_lighter.sc
@@ -57,7 +57,7 @@ spread_torches(player, initial_gamemode) ->
 		loop(4000,
 			lpos = cpos+l(rand(d), rand(d), rand(d)) - d/2;
 			if (__distance_sq(cpos, lpos) <= dd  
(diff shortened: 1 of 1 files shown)


altrisi (approved):
Habemus torches! It seems to work in my testing!

PR #322, opened 2022-06-04
Block breaker

DragonCrafted87:
New app that combines and extends the silk_spawners and silk_budding_amethyst scripts


programs/survival/silk_breakables.sc
```diff
@@ -0,0 +1,238 @@
...
+        mining_fatigue_level == 2,
+        (
+            speedMultiplier = speedMultiplier * 0.0027;
+        ),
+        // else if
+        mining_fatigue_level >= 3,
+        (
+            speedMultiplier = speedMultiplier * 0.00081;
+        )
+    );
```
opsaaaaa, line 114: I like the `else if` comments btw.

I would tend to attempt to use a map/hash for this. 
Tho I see that could also get complicated.

opsaaaaa (approved):
Budding amethyst drops inconsistently. Im uncertain if this is intended.
I doesn't seem to drop at all with netherite pickaxe.

perhaps the hardness is less then default.
> [hardness] must be less than the default ... or there's a good chance that the drop ... won't trigger

Everything else i tested works fine tho.

Good app. generalizing silk-able blocks is really good.
And the code quality is nice.

DragonCrafted87:
> Budding amethyst drops inconsistently. Im uncertain if this is intended. I doesn't seem to drop at all with netherite pickaxe.
> 
> perhaps the hardness is less then default.
> 
> > [hardness] must be less than the default ... or there's a good chance that the drop ... won't trigger
> 
> Everything else i tested works fine tho.
> 
> Good app. generalizing silk-able blocks is really good. And the code quality is nice.

yeah thats one thing i haven't been able to fully resolve if you do them one at a time it should drop them all but if you do something like a tower stack it'll somewhat inconsistently revert to triggering the vanilla block break

programs/survival/silk_breakables.sc
DragonCrafted87: thanks i needed to find some way to make it readable as i've been working on this one off and on over a couple of months and since my day to day is c++ & python it takes a bit for me to switch to scarpets more pure functional syntax w/o comments like that

Ghoulboy78 (approved):

rv3r (commented):
Script works as described for, at minimum, the single case of netherite pick, haste 2, efficiency 5, silk touch. Working on checking other combinations. Spawner placing works as intended. Probably will approve, just not yet.

PR #323, opened 2022-06-12
added cart_breaker app

nsluhrs:
Adding cart_breaker app which adds the ability when holding an axe and sneak left clicking a combined minecart (Minecart with chest, TNT etc) or boat with chest will drop inventory and the component parts of the combined entity as items.

### DIFF
--- a/programs/survival/cart_breaker.sc
+++ b/programs/survival/cart_breaker.sc
@@ -1,0 +1,30 @@
+// cart_breaker.sc 
+// by kergadon aka nsluhrs
+// Breaks minecarts such as chest minecarts or hopper minecarts also boats with chests
+__config()->{'stay_loaded'->true};
+
+__on_player_attacks_entity(player, entity) ->(
+  mainhand=query(player,'holds','mainhand');
+  // checks that an axe is held and player has right clicked
+  if(query(player, 'sneaking') && mainhand:0 ~ '_axe',
+    entity_name=str(entity);
+    if(entity_name ~ 'Minecart with ' || entity_name ~ 'Boat with Chest',
+      // drops all items in inventory 
+      loop(inventory_size(entity), drop_item(entity, _));
+      ep = pos(entity);
+      // removes cart
+      modify(entity,'remove');
+      if(entity_name~'Minecart',
+        // makes it so dropped items fall a short distance      
+        ep:1 =ep:1+0.5;
+        // cart case
+        spawn('item', ep, {'Item' -> {'id' -> 'minecart', 'Count' -> 1}});
+        spawn('item', ep, {'Item' -> {'id' -> entity~'type'~'(.+)_', 'Count' -> 1}}),
+        // boat chest
+        boat_type=query(entity,'nbt','Type');
+        spawn('item', ep, {'Item' -> {'id' -> boat_type+'_boat', 'Count' -> 1}});
+        spawn('item', ep, {'Item' -> {'id' -> 'chest', 'Count' -> 1}});
+      );
+    );
+  );
+);
--- a/programs/survival/README.md
+++ b/programs/survival/README.md
@@ -51,6 +51,10 @@ Various scripts that modify various game elements, often replicating popular mod
 #### By KingOfTheClouds
 	A Totem of Undying anywhere in the inventory (except in shulker boxes etc.) can activate as if it were held.
 
+### [cart_breaker.sc](https://github.com/gnembon/scarpet/blob/master/programs/survival/cart_breaker.sc)
+#### By nsluhrs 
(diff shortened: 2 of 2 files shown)


programs/survival/cart_breaker.sc
```diff
@@ -0,0 +1,30 @@
...
+// Breaks minecarts such as chest minecarts or hopper minecarts also boats with chests
+__config()->{'stay_loaded'->true};
+
+__on_player_attacks_entity(player, entity) ->(
+  mainhand=query(player,'holds','mainhand');
+  // checks that an axe is held and player has right clicked
+  if(query(player, 'sneaking') && mainhand:0 ~ '_axe',
+    entity_name=str(entity);
+    if(entity_name ~ 'Minecart with ' || entity_name ~ 'Boat with Chest',
+      // drops all items in inventory
```
opsaaaaa, line 12: I don't really like checking for things using the name of an item/entity. 
I this will return a different result depending on the language settings.
So this app as is won't work when I set my language settings to Spanish for example.
ie. `Vagon con TNT`

instead you could query the `'type'` which returns the same thing regardless of language.

```
entity~'type'
tnt_minecart
```

opsaaaaa (changes requested):
I get that this is intended for `1.19` however I tested it on `1.18.1`. I don't really want to update all my mods rn. This made testing kinda pointless as its default behavior for this version.

I didn't get errors when I loaded the app tho. 

It probably won't work with language settings other than English.
As its using the entity name to check if the items type of minecart.

Other than that, pretty good code.

programs/survival/cart_breaker.sc
```diff
@@ -0,0 +1,30 @@
...
+    entity_name=str(entity);
+    if(entity_name ~ 'Minecart with ' || entity_name ~ 'Boat with Chest',
+      // drops all items in inventory 
+      loop(inventory_size(entity), drop_item(entity, _));
+      ep = pos(entity);
+      // removes cart
+      modify(entity,'remove');
+      if(entity_name~'Minecart',
+        // makes it so dropped items fall a short distance      
+        ep:1 =ep:1+0.5;
```
SurfingDude-1182, line 19: Minor nitpick (aside from what opsaa said), the straight drop here is a bit jarring, it might be aesthetically pleasing to add a small random velocity to emulate vanilla-like drop behavior.

programs/survival/cart_breaker.sc
```diff
@@ -0,0 +1,30 @@
...
+// by kergadon aka nsluhrs
+// Breaks minecarts such as chest minecarts or hopper minecarts also boats with chests
+__config()->{'stay_loaded'->true};
+
+__on_player_attacks_entity(player, entity) ->(
+  mainhand=query(player,'holds','mainhand');
+  // checks that an axe is held and player has right clicked
+  if(query(player, 'sneaking') && mainhand:0 ~ '_axe',
+    entity_name=str(entity);
+    if(entity_name ~ 'Minecart with ' || entity_name ~ 'Boat with Chest',
```
Ghoulboy78, line 11:
```suggestion
    if(entity_name ~ '_minecart' || entity_name ~ '_chest_boat',
```

programs/survival/cart_breaker.sc
Ghoulboy78: Suggested a change to fix this

programs/survival/cart_breaker.sc
Ghoulboy78: This is to fix the issue raised by opsaa

Ghoulboy78, line 11:
```suggestion
    entity_type=entity~'type';
    if(entity_type~ '_minecart' || entity_type~ '_chest_boat',
```

This is to fix the issue raised by opsaa

altrisi:
It's always dropping a chest for me for any composed minecart type, And boats with chest aren't being affected at all.

programs/survival/cart_breaker.sc
```diff
@@ -0,0 +1,36 @@
...
+// Breaks minecarts such as chest minecarts or hopper minecarts also boats with chests
+__config()->{'stay_loaded'->true};
+
+__on_player_attacks_entity(player, entity) ->(
+  mainhand=query(player,'holds','mainhand');
+  // checks that an axe is held and player has right clicked
+  if(query(player, 'sneaking') && mainhand:0 ~ '_axe',
+    entity_type=entity~'type';
+    //checks if entity is seprable
+    if(entity_type~ '_minecart' || entity_type~ 'chest_boat',
```
Ghoulboy78, line 12:
```suggestion
    if(entity_type~ '_minecart' || entity_type == 'chest_boat',
```
We can replace with this ig

altrisi (approved):
Looks and works great!

Ghoulboy78 (approved):
Agree with altrisi, this is ready to merge!

altrisi:
Thanks for contributing!

PR #326, opened 2022-06-27
Adding light_level_overlay.sc

Ghoulboy78:
Cos MiniHUD wasn't updated but I like this functionality, so I decided to add it myself. It has some utility features too for busy servers and poor connections.

### DIFF
--- a/programs/survival/README.md
+++ b/programs/survival/README.md
@@ -133,6 +133,12 @@ Various scripts that modify various game elements, often replicating popular mod
 #### By _GieR
     Make sponge work on lava like water.
 
+### [light_level_overlay.sc](https://github.com/gnembon/scarpet/blob/master/programs/survival/light_level_overlay.sc):
+#### By Ghoulboy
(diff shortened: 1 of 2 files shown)


rv3r:
Mob spawning light level doesn't work atm gnembon/fabric-carpet#1655

rv3r:
Unless your above task means something different than the `range` option that's already present, I feel like this is good to go.

Ghoulboy78:
I mean only putting the display on blocks that the player can view, cos otherwise it gets laggy af, even for small ranges.

rv3r:
> I mean only putting the display on blocks that the player can view, cos otherwise it gets laggy af, even for small ranges.

Isn't that basically what the armor stand combined with `'trace'` does?

Ghoulboy78:
Yes, but this solution is even laggier

rv3r:
Oh good point. Plus it requires a lot of subdivisions to look nice, which makes things worse.

PR #328, opened 2022-07-11
Horse Buddy App

opsaaaaa:
The goal of this app is to make getting a fast horse more fun.
And to remove the 'need' to slay tons of horses to stop the lag.

The original idea was to add a mechanic inspired by villager cycling (zombifing and curing villagers to get cheep trades).
Something that used potions in a similar way. But hopefully less brutal.

The app drifted a way from that a little.
The horse has a preferred food at any one time.
feeding the horse the food its craving can trigger a new stat roll.
The roll adjusts the stats of the horse by a hopefully reasonable amount.

You can improve the odds in two ways.
One by ridding that horse a lot. Only really effective if you stick to one horse.
You can also use potions like swiftness or jump_boost to bias the roll towards one or the other. And strength potions just improve the chances overall. 
All the effect accumulate. 
so you can stack the strength and swiftness for example.

There needed to be a way to communicate the mechanics to players.
So I kinda made the horses complain about their food.
Which sounds annoying but is actually pretty amusing.

The dialog feels a kinda unminecrafty. but I don't see players discovering the mechanics without it. 

(aaa, don't judge the dialog to much.)


programs/survival/horse_buddy.sc
```diff
@@ -0,0 +1,306 @@
...
+_rand_message(p,code,num) -> (
+  _message(p,str('%s.%d', code, floor(rand(num + 0.9))));
+);
+
+
+//--- LOCALS ---//
+
+_i18n(code) -> ({
+  'en'->{
+    'random.hint.0'       -> 'giv\'me jumpy jump magic!',
```
Ghoulboy78, line 273: The text is adorable, and totally untranslatable into any of the languages I know. I love it.

programs/survival/horse_buddy.sc
```diff
@@ -0,0 +1,306 @@
...
+//--- ON PLAYER METHODS ---//
+
+__on_player_rides(p, forward, strafe, jumping, sneaking) -> (
+  global_training:'horse.jump_strength' += number(jumping);
+  global_training:'generic.movement_speed' += ceil(forward) + ceil(strafe);
+);
+
+__on_player_interacts_with_entity(p, horse, hand) -> (
+  kind = horse~'type';
+  if(kind == 'horse' || kind == 'donkey',
```
Ghoulboy78, line 177: does this exclude mules? Or are they just classed as donkeys?

programs/survival/horse_buddy.sc
```diff
@@ -0,0 +1,306 @@
...
+);
+
+//--- ROLL NEW STATS ---//
+
+_roll_new_stats(horse) -> (
+  effect_mod = _effect_stat_mods(reduce(horse~'effect', _a += _:0, []));
+  attributes = horse~'attribute';
+
+  // these are the magic numbers that try to balance out the training
+  // we need to turn numbers like 500 and 6000 into reasonable stat modifiers no larger than 1.
```
Ghoulboy78, line 227: Can u figure out the largest possible value these numbers could have? If so you could take the log and then scale them (so smaller numbers also have some significance at that scale)
If not, just do log (base 10) and divide by like 4 (cos that way 10,000 becomes 1 and 10 becomes 0.25)

programs/survival/horse_buddy.sc
opsaaaaa: lol, it's not like it needs to be 1-1 translation! just like basically, horse puns and stuff, with whatever horses sound like on paper in that language. Every language has horse puns! probably.

programs/survival/horse_buddy.sc
opsaaaaa: added mules

programs/survival/horse_buddy.sc
opsaaaaa: Ok I tired the log stuff. The problem is I actually just want the effort to be linear.
Any exponential effects I would want to be slight.
And getting exponential math to feel right in game is more effort.

opsaaaaa:
I did a less terrible job on the training math.
and I added mule support.

PR #329, opened 2022-07-11
Rope ladder update

opsaaaaa:
I wanted to try this app on my server, sorry I kinda rewrote the whole thing.

notable changes:
- new feature: unsupported ladders rope ladders break. 
- less strict about the face that was clicked on.
- removed `property` and replaced with `block_state`
- refactored ~~a lot~~ *everything*. *(probably to a pointless extent)*


programs/survival/rope_ladders.sc
```diff
@@ -4,27 +4,88 @@
...
 
-__on_player_right_clicks_block(player, item_tuple, hand, block, face, hitvec) -> (
-    _rope_ladders(player, item_tuple, hand, block, face)
-)
+
+_loop_to_ladder_base(block) -> (
+    b = _under_ladder(block);
+
+    if(b == 'ladder' && block_state(b,'facing') == block_state(block, 'facing'),
+
```
opsaaaaa, line 54: This is to prevent the rope ladder from continuing a ladder below itself in the wrong orientation.
I could live without the block state check tbh.

programs/survival/rope_ladders.sc
```diff
@@ -4,27 +4,88 @@
...
+    );
+    return(b)
+);
+
+
+_attempt_place_ladder(base, root) -> (
+  if(air(base) || base == 'water', 
+    set(base, root, 'waterlogged', if(base == 'water', 'true', 'false'));
+  );
+);
```
opsaaaaa, line 66: I personally don't care if rope ladders can extend into the water, but I was in the original app.

programs/survival/rope_ladders.sc
```diff
@@ -4,27 +4,88 @@
...
-    inventory_set(player, if(hand=='mainhand',player~'selected_slot',-1), count - 1, item ,nbt)
+  );
+);
+
+
+_rope_should_start(p, item, block) -> (
+    p~'gamemode' != 'spectator' &&
+    item && item:0 == 'ladder' &&
+    block~'ladder' &&
+    if(g=='adventure', nbt:'CanPlaceOn'~'"minecraft:ladder"', true)
```
Ghoulboy78, line 46: What is the variable `g`?

programs/survival/rope_ladders.sc
Ghoulboy78: Do you think it would be possible to do this without recursion? Cos like imo a while loop fits here just fine, and so you can also add a limit, like the world height, in case something goes wrong.

programs/survival/rope_ladders.sc
```diff
@@ -4,27 +4,88 @@
 // (Carpet Mod 1.4.9)
 //
 // Right clicking on a ladder with an other one, will extend the existing one down.
-///
+
+// 'Updated' by opsaaaaa for 1.19
+// added break unsuported ladder fearue
```
Ghoulboy78, line 9: I think it ought to say something along the lines of `by BisUmTo and opsaaa`, cos he forgot to add his own name.

programs/survival/rope_ladders.sc
opsaaaaa: A while loop would likely work. I'll try it.

programs/survival/rope_ladders.sc
opsaaaaa: Aww it's from the original script, it was the player game mode. I think I removed the definition at some point.

opsaaaaa:
@Ghoulboy78 thanks for looking at this. 
I'll add some fixes probably sometime this week.

opsaaaaa:
@Ghoulboy78

I agree that some players probably liked the unsupported ladders.
And come to think about it, its not actually what i wanted anyway.
So I propose something else instead.

> As a player standing next to a rope ladder.
> I want to be able to dismantle the rope ladder as easily as I deployed it.

> Given that I am holding shift
> When I break a ladder
> It should also break vertically connected ladders.

opsaaaaa:
I also added 'easy_pickup` and `sky_ropes`.

> Given I am holding shift
> And my ladder has bottomed out
> When I place the ladder
> Then it should extend the ladder into the sky.

> As a player at the top of a cliff with a rope hanging down
> When I dismantle the ladder
> Then It the ladder items should teleport to me

opsaaaaa:
I found and fixed a bug with useup_item function.
It was setting the first hotbar slot instead of the selected_slot.
Its working correctly now, and I added the place ladder sound.

altrisi:
Only one thing from my testing, my ladders float after breaking supporting ones, unless sky_ladders is off.

opsaaaaa:
@altrisi 
I'll be totally honest. I cant guarantee I'll get to fixing that. Probably not tbh. If i do, it will be a few months.

altrisi:
Makes sense, we've been exceptionally slow (well, sadly honestly not sure if exceptionally). I'll test it back later and if it's not a regression I'll approve it, and just make an issue for improvement, that way it'll only be missing one review, which ig is better.

altrisi (approved):
From further reading looks like it was an intended change, when adding the sky ropes feature, it's just the description wasn't edited.

opsaaaaa:
I updated the description to reflect the changes. I should have done that initially.
 
I've also been running this on a server with friends for like 8-10 months now. No real issues with it. So its had lots of in game play testing.

I think this could probably be easily merged. unless there is some other issue with it.

@Ghoulboy78

programs/survival/README.md
```diff
@@ -254,10 +254,14 @@ Various scripts that modify various game elements, often replicating popular mod
...
-#### By BisUmTo
-	Right clicking on a ladder with an other one, will extend the existing one down.
+#### By BisUmTo and Opsaaaaa
+	Right clicking on a ladder with another one, will extend the existing one down.
 	https://youtu.be/Mgz6XEIcpgc
 
+  `easy_dismantle` while holding shift destroys connected ladders
+  `easy_pickup` will teleport nearby ladder items to the player
+  `sky_ropes` allows you to extend ropes into the sky while holding shift
+
```
Advait-Sen, line 264:
```suggestion
	https://youtu.be/Mgz6XEIcpgc

	`easy_dismantle` while holding shift destroys connected ladders
	`easy_pickup` will teleport nearby ladder items to the player
	`sky_ropes` allows you to extend ropes into the sky while holding shift
```

Ghoulboy78 (approved):
Just updated Readme formatting a little, but looks good otherwise!

Ghoulboy78:
Also, if  you got a notification from another account, that's cos I was trying to review while logged onto another account, so oops...

PR #331, opened 2022-07-21
Add single bedrock app

ysc3839:
Find single bedrock on top of the nether.
Just like the "Bedrock Ceiling" feature of [BoundingBoxOutlineReloaded](https://github.com/irtimaled/BoundingBoxOutlineReloaded).
Type `/single_bedrock` to turn on rendering, and type again to turn off.
It checks 9x9 chunks around the player on every 20gt.
It seems that carpet shapes can be occluded by blocks. (Is there an option to make shapes always visible behind blocks?)
So if the boxes are occluded, type `/single_bedrock bottom` to make it from y=0 to y=127, type again to turn off.

Tested on 1.18.2.

Screenshots:


SurfingDude-1182:
The app works great, tested it on 1.17.1. I would still like to provide suggestions for some enhancements,
1. Since this was inspired by BBOR, having a rendering system similar to it would be much more user friendly/short than this one imo. BBOR makes all boxes, boundaries, etc. based on the player's Y level (or it used to at least). So what I am proposing is, check the player's Y position and draw box from there to the y-128 location. With the current system, everyone almost always has to use the bottom sub command if they are below nether roof. `Bottom` will still retain its use case, but when a player wants something quick, BBOR style of rendering might be easier.

Everything following this are minor nitpicks,

2. `'_tick'` runs even if the player is not moving `'on_move'` event might come in handy here.
3. `Enabled` and `Disabled` depends on gamerule `sendCommandFeedback`. Traditional print might be better for servers that have it disabled.
4. After removing all bedrock from a column, the box still persists. While this does not affect much on a smaller scale, if the `draw_radius` is increased and someone goes near a mob farm with roof broken in a massive area. I think the large amount of boxes can cause some performance issues.
5. I don't know enough about `task()` but I think it would help a lot, if someone increases the draw radius to a big number.
6. Afaik `strict` does not really serve any purpose here.
7. I was confused with all the 'constants' being stored as a variable, looking at `chunk_size` & `block_name` specifically.

Regardless of these nitpicks, the app works great with my limited testing.

ysc3839:
@SurfingDude-1182 Thank you for your suggestions!
1. BBOR renders only one block and can be occluded by blocks: 
2. I think `on_move` would trigger more frequent. It's a good idea to check player's chunk before calculating.
3. I will change it.
4. I will change it.
5. I think using `task()` would not increasing MSPT, because `Scheduled functions run at the end of the tick` ([link](https://github.com/gnembon/fabric-carpet/blob/master/docs/scarpet/Full.md#scheduledelay-function-args)). I will try using `task()` to see if it's better.
6. I will change it.
7. It seems that there's no constants system in carpet script, so I use global variables as constants.

ysc3839:
It's very slow when using `task()`. I set `DRAW_RADIUS` to 4, with task it takes 2138ms, without task only 73ms.

ysc3839:
Updated.
Add shapes cache. It will only calculate when the player move to another chunk.

programs/survival/single_bedrock.sc
```diff
@@ -0,0 +1,82 @@
...
+global_last_chunk_z = null;
+global_shapes = null;
+
+__command() -> (
+  global_player = player();
+  global_enabled = !global_enabled;
+  _clear_cache();
+  if(global_enabled, (
+    in_dimension(global_DIM, schedule(0, '_tick'));
+    print('Enabled');
```
Ghoulboy78, line 29: Do you think this could be replaced by an actionbar message? Cos that way it only goes to the player that's using it, and no one else.

programs/survival/single_bedrock.sc
```diff
@@ -0,0 +1,82 @@
...
+_find_in_chunk(chunkX, chunkZ, shapes) -> (
+  chunkStartX = chunkX * global_CHUNK_SIZE;
+  chunkStartZ = chunkZ * global_CHUNK_SIZE;
+  c_for(x = chunkStartX, x < chunkStartX + global_CHUNK_SIZE, x += 1,
+    c_for(z = chunkStartZ, z < chunkStartZ + global_CHUNK_SIZE, z += 1,
+      if(block(x, global_TOP_Y, z) == global_BLOCK_NAME &&
+        all(rect([x, global_TOP_Y - 1, z], [0, global_CHECK_HEIGHT - 1, 0], [0, 0, 0]), _ != global_BLOCK_NAME),
+        from = [x, global_TOP_Y - global_DRAW_HEIGHT, z];
+        if(global_from_bottom, from:1 = 0);
+        shapes += ['box', global_DURATION, 'from', from, 'to', [x + 1, global_TOP_Y + 1, z + 1], 'color', global_COLOR, 'fill', global_COLOR];
```
Ghoulboy78, line 78: Same here, you can add `'player', global_player` so it only goes to that player.

programs/survival/single_bedrock.sc
```diff
@@ -0,0 +1,82 @@
...
+global_INTERVAL = 20;
+global_DURATION = global_INTERVAL + 2;
+global_TOP_Y = 127;
+global_CHECK_HEIGHT = 4;
+global_BLOCK_NAME = 'bedrock';
+global_DRAW_HEIGHT = 6;
+global_COLOR = 0x00ff0050;
+
+__config() -> {
+  'command_permission' -> 'players'
```
Ghoulboy78, line 13: Just out of curiosity, what would happen if you were to get rid of this line?

ysc3839: I didn't test on a server. I think execute in server console would make `player()` return `null`. So I add this to prevent execute from console.

programs/survival/single_bedrock.sc
ysc3839: Looks like other players would not see that message. I tested using LAN multiplayer.

SurfingDude-1182:
> Updated. Add shapes cache. It will only calculate when the player move to another chunk.

Yep, this is exactly what I had in mind.

programs/survival/single_bedrock.sc
Ghoulboy78: wait, do you want other players to be able to see this message?

ysc3839: No, I don't.

opsaaaaa (approved):
I tested it and look at the code.
It looks good! and It works well.

Wonderful app, Looking forward to using it while playing.

programs/survival/single_bedrock.sc
altrisi: Print without arguments only goes to the executing source.

altrisi (approved):
Works great!

PR #335, opened 2022-07-30
Tputils: Home and Warps

SurfingDude-1182:
All actions can be performed with CLI (like the waypoints app). Additionally, GUI for actions like displaying/teleporting to available warps. The GUI should only appear if the carpet version > 1.4.56 (I don't remember testing this, but I hope it works) so no need to bump up required version.

### DIFF
--- a/programs/survival/tputils.sc
+++ b/programs/survival/tputils.sc
@@ -1,0 +1,283 @@
+global_max_homes = 20; //Edit to allow more/less homes
+global_operator_permission_level = 2;
+
+__config()->{
+    'scope' -> 'global',
+    'commands' ->
+    {   
+        'home add <nameh>' -> '_hadd',
+        'home tp <hlocation>' -> '_htp',
+        'home remove <hlocation>' -> '_hremove',
+        'home list' -> '_hlist',
+        'warp add <namew> <description>' -> '_wadd',
+        'warp tp <wlocation>' -> '_wtp',
+        'warp remove <wlocation>' -> '_wremove',
+        'warp menu' -> '_wmenu'
+    },
+    'arguments' ->
+    {
+        'hlocation' -> {
+            'type' -> 'text',
+            'suggester' -> _(args) -> (
+                keys(global_hdata:str(player()~'uuid'))
+            ),
+            'case_sensitive' -> false
+        },
+
+        'nameh' -> {
+            'type' -> 'text',
+            'suggest' -> ['Spawn','Base','End Portal','Nether Portal'],
+            'case_sensitive' -> false
+        },
+
+        'namew' -> {
+            'type' -> 'string',
+            'suggest' -> ['Spawn','Marketplace','Arena'],
+            'case_sensitive' -> false
+        },
+
+        'description' -> {
+            'type' -> 'text'
+        },
+
+        'wlocation' -> {
+            'type' -> 'text',
+            'suggester' -> _(args) -> (
+                keys(global_wdata)
+            ),
+            'case_sensitive' -> false
+        }
+    }
+};
+
+
+//Initializing homes and warps data
+
+if(read_file('hdata','JSON') == null,
+    global_hdata = {},
+    //else when file already exists
+    global_hdata = read_file('hdata','JSON');;
+);
+if(read_file('wdata','JSON') == null,
+    global_wdata = {},
+    //else when file already exists
+    global_wdata = read_file('wdata','JSON');;
+);
+
+
+
+//Home functions
+
+_hadd(name) -> (
+    
+    player=player();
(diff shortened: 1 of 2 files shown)


programs/survival/tputils.sc
```diff
@@ -0,0 +1,283 @@
...
+    if(global_hdata:uuid==null,global_hdata:uuid={});
+    global_names = keys(global_hdata:uuid);
+    if(global_names~name!=null, print(player,format(str('r %s is already in use. Therefore, updating the old home to new location. \n',name))));
+
+    global_hdata:uuid:name={};
+    global_hdata:uuid:name:'location' = pos(player);
+    global_hdata:uuid:name:'pitch' = player~'pitch';
+    global_hdata:uuid:name:'yaw' = player~'yaw';
+    global_hdata:uuid:name:'dimension' = player~'dimension';
+    _saveHData();
```
Ghoulboy78, line 89: Same as with another pr u've done I think, I think you should do saveData in __on_close() and that's it to avoid re-writing to disk a bunch of times

programs/survival/tputils.sc
```diff
@@ -0,0 +1,283 @@
...
+);
+
+_screenCheck() -> (
+
+    sVersion= split('\\+',system_info('scarpet_version')):0; //fetches first part before the +, eg 1.4.57
+    vNumbers=split('\\.',sVersion);
+    if(number(vNumbers:0)>=1 && number(vNumbers:1)>=4 && number(vNumbers:2)>=57,
+        return (true);,
+    //Cant make screen
+        return(false);
```
Ghoulboy78, line 281:
```suggestion
    number(vNumbers:0)>=1 && number(vNumbers:1)>=4 && number(vNumbers:2)>=57
```
You can avoid return() statements, cos this does the same and also return() statements are pretty bad (for some reason, I still don't know)

programs/survival/tputils.sc
```diff
@@ -0,0 +1,283 @@
...
+            for(keys(global_wdata),
+
+                info = global_wdata:_;
+                iType=info:'item';
+                if(iType==null,iType='barrier');
+                des=info:'description';
+                loc=info:'location';
+                x= floor(loc:0);
+                y= floor(loc:1);
+                z= floor(loc:2);
```
Ghoulboy78, line 240: This is a completely unrelated point, but this is the same as `[x, y, z] = map(loc, floor(_))`.

SurfingDude-1182: Oman my whole scarpet life was a lie. I never knew this was possible xD. Imma change it everywhere asap

programs/survival/tputils.sc
SurfingDude-1182: Yh imma change this

programs/survival/tputils.sc
```diff
@@ -0,0 +1,279 @@
...
+    global_hdata:uuid:name:'yaw' = player~'yaw';
+    global_hdata:uuid:name:'dimension' = player~'dimension';
+    print(player,format('e Succesfully saved home to location with name: '+name));
+);
+
+_htp(name) -> (
+
+    info = global_hdata:str(player()~'uuid'):name;
+    if(info==null,
+        print(player,format('r \nThe specefied location does not exist. Please use home list to know your current list'));
```
altrisi, line 104:
```suggestion
        print(player(),format('r \nThe specefied location does not exist. Please use home list to know your current list'));

```

programs/survival/tputils.sc
```diff
@@ -0,0 +1,279 @@
...
+    yaw = info:'yaw';
+    location = info:'location';
+
+    run(str('execute in %s run tp %s %f %f %f %f %f',dimension,player(),location:0,location:1,location:2,yaw,pitch));
+
+);
+
+_hremove(name) -> (
+    if(global_hdata:str(player()~'uuid'):name==null,
+        print(player,format('r \nThe specefied location does not exist. Please use home list to know your current list'));
```
altrisi, line 118:
```suggestion
        print(player(),format('r \nThe specefied location does not exist. Please use home list to know your current list'));

```

programs/survival/tputils.sc
```diff
@@ -0,0 +1,279 @@
...
+  main == 1,
+  if(sub>4,
+    true,
+    sub==4,
+    x >= 56,
+    false
+  ),
+  false
+)
+);
```
altrisi, line 279: Suggestion for if this is false just don't register the subcommand, else a command that does nothing is weird.

altrisi (commented):
Found some bugs. Also I think it'd be nicer if there was a message (maybe to actionbar?) when teleporting, so you know something has happened, though it's not really needed.

Issue #311, opened 2022-04-22
Teleport to safe spot, but as a library


altrisi:
`world_map` has a function to teleport to a safe spot:

https://github.com/gnembon/scarpet/blob/022a0a7700693c3dcd8b4190e492ff8fa66c3e44/programs/survival/world_map.sc#L1547-L1564

`cam` has something for that too:

https://github.com/gnembon/scarpet/blob/8c9c4a56e656a3169a3420be058c60601d4a9e31/programs/survival/cam.sc#L156-L163

And the thing is, it's likely other apps would like to use similar functionality, such as waystones in #305.

Proposal for extracting it to a library. Maybe not making cam use the library as cam is used a lot and adding a dependency could be an issue, but I don't see why world_map can't, same for other apps.

Ghoulboy78:
Well, then would this entail potentially creating a block in midair tho? cos thats what world_map does, but not everyone want to do that. Unless we do it so that it returns null if there is no safe space in the area they query.

altrisi:
Well if you opt-in to the library, it will do it, but the lib would tell you what it does. Could always have a param or something to decide whether it should create a platform if needed.

PR #336, opened 2022-07-30
Graves: Inspired by Universal Graves

pseudofractal:
Its almost an 1 to 1 copy of [Universal Graves mod](https://github.com/Patbox/UniversalGraves) except the fancy holograms and some minor stuff. Since this uses fake screens, the minimum carpet version is bumped up to 1.4.57.

Spent a lot of time trying to make things as user friendly as possible. I am not very happy with the stuff I am doing to find a "safe" grave location, but I think the current solution is the best until #311.

Another thing, I am using player Name in all places instead of uuid because I did not find any good way of fetching the uuid from a name (Player heads store the player name and not uuid in their block data and Mojang api is rate limited)

Another another thing, I am using both `set()` and `run('setblock')`  in different places. The reason for this is, `set` sometimes fails (particularly when replacing a player head) I need to do some more tests and then make a issue regarding this.


programs/survival/graves.sc
```diff
@@ -0,0 +1,349 @@
...
+    'stay_loaded' -> true,
+    'scope' -> 'global',
+    'requires' -> {
+        'carpet' -> '>=1.4.57'
+    }
+};
+
+__command() -> null;
+
+gdata_file = read_file('gdata','JSON');
```
Ghoulboy78, line 19: Wouldn't it be easier to use `store_app_data()` and do it in NBT? unless you want JSON specifically?

programs/survival/graves.sc
```diff
@@ -0,0 +1,349 @@
+// Tweak Stuff
```
Ghoulboy78, line 1: You can add your name as well as a small description up at the top

programs/survival/graves.sc
```diff
@@ -0,0 +1,349 @@
...
+            );      
+        );
+);
+
+
+
+
+//Utility Functions
+
+_request(global_gdata,block) -> (
```
Ghoulboy78, line 232: if this is a global variable, then why do you need to pass it as an argument? Surely you can just grab it from within the function? (unless there is a specific reason for passing it in as an argument ofc)

SurfingDude-1182: This is plain old embarrassing xD, there is no reason. I overlooked this while going through the thing, also I dont even think its necessary to put this one-liner into a separate function. Imma change this too

programs/survival/graves.sc
SurfingDude-1182: Aye will do

programs/survival/graves.sc
SurfingDude-1182: When I first started learning about data storage concepts, my reference was waypoints app. Since it has to store multiple files, ig firi used this method. I sorta never stopped using this after learning from it. If there is a difference in performance, Imma go ahead and change it.

Ghoulboy78: store_app_data is the most performance-effective way, as well as taking up less space to store the info

programs/survival/graves.sc
```diff
@@ -0,0 +1,419 @@
...
+if(settings_file == null,
+    global_settings = {
+        'requestTime' -> 30,
+        'allowRequests' -> true,
+        'allowOnlyOwner' -> false,
+        'operatorPermissionLevel' -> 2
+    },
+    //else when settings exist
+    global_settings = settings_file;
+);
```
Ghoulboy78, line 61:
```suggestion
app_data = parse_nbt(load_app_data());

if(has(app_data, 'gdata'),
    global_gdata = app_data:'gdata',
    //else if data is not present
    global_gdata = {}
);

if(has(app_data, 'setings'),
    global_settings = app_data:'setings',
    //else if data is not present
    global_settings = {
        'requestTime' -> 30,
        'allowRequests' -> true,
        'allowOnlyOwner' -> false,
        'operatorPermissionLevel' -> 2
    }
);
```

This change allows to read code using `load_app_data()`

programs/survival/graves.sc
```diff
@@ -0,0 +1,419 @@
...
+    c_for(i=0,i<inventory_size(player),i+=1,
+
+        if(inventory_get(player,i)!=null,
+
+            slotItem= inventory_get(player,i);
+            global_gdata:str(gravepos):str(i) = slotItem;
+            item_count+=1;
+            _saveGData();
+        
+        );
```
Ghoulboy78, line 151: Couldn't you do this as a list? Like wouldn't that make it easier to iterate through?

programs/survival/graves.sc
```diff
@@ -0,0 +1,419 @@
...
+
+);
+
+_delete_data_slot(data,slotNum) -> (
+
+    delete(data,str(slotNum));
+    delete_file('gdata','JSON');
+    _saveGData();
+
+);
```
Ghoulboy78, line 328:
```suggestion
__on_close()->(
    app_data = {
        'gdata'->global_gdata,
        'settings'->global_settings
    };
    store_app_data(encode_nbt(app_data, false))
);
```

TBH you can get rid of these functions (and all their mentions), cos you only rly need to save to disk in the `__on_close()` event which fires right as the server is closing (still test it tho ofc).

Note, change the false to true if it gives you trouble in saving, though to see that you will probably need the output log open. It's just cos im not sure if it will accept the way you store the player data. If it doesn't, I'll help you with it dw.

Ghoulboy78 (commented):
Here is some feedback, plus the code required to move to `store_app_data`. It may a be a bit of work, but honestly I think it's an improvement over saving 2 json files over and over again in the code.

programs/survival/graves.sc
SurfingDude-1182: I did not do this as a list because the grave auto-pickup by owner feature depends on being able to know the exact position of each item in inventory. I had to add the inventory==null check just to get around the issue of a very big save file

altrisi:
Found an issue, if you drown the head will pop, and bye items.

altrisi:
Also curious what determines if the inventory will be restored directly or if the screen will open?

altrisi:
> Found an issue, if you drown the head will pop, and bye items.

Actually, the head will keep some grave data, so if you place it and right click/break it it will disappear, but the items won't be restored.

PsuedoFractal:
> Also curious what determines if the inventory will be restored directly or if the screen will open?

Well two things. 
1. If the clicker is the grave owner, then it will try to shift all items directly to inventory and try to maintain positions of items. If its someone else, then they get the screen.
2. While trying to shift items, if the inventory already contains some item in slot x, then it will open the screen.

PsuedoFractal:
> Found an issue, if you drown the head will pop, and bye items.

Ahh I never tried drowning but it does make sense. The debug command would fix this but a better solution is needed yh....

gnembon (approved):

Issue #341, opened 2022-09-22
Camera app won't install


unofficialreece:
**Describe the bug**
The camera module app won't install because there is an error (Error while evaluating expression). More information about the errors below.

**To Reproduce**
1. Type the command `/script download utilities/camera.sc`
2. An error occurred.

**Expected behavior**
The Camera app is supposed to be installed properly with no errors.

**Screenshots**
![2022-09-23_02 26 03](https://user-images.githubusercontent.com/79681820/191834446-dff29dcd-f42d-4581-b8c7-b309cb00aba7.png)

**Environment**
 - Carpet version: 1.4.83
 - Carpet Extra version (if needed): 1.4.83
 - Minecraft version: 1.19.2
 - Quilt version: 0.17.5-beta.2

altrisi:
That camera app is now built into Carpet, you can load it with just `/script load camera` (after removing the broken copy from your scripts folder).

Note that `camera` is the app to make camera paths and stuff, if you want the one for going into spectator mode it's in `survival/cam.sc`.

PR #188, opened 2021-02-24
Update camera.sc compared to Fabric Carpet (alternative to #187)

DragonEggBedrockBreaking:
I made #187 to remove the script, but then @BisUmTo said we should just update it. I will keep both PRs open and you can choose which to merge.

PR #343, opened 2022-10-08
Delete camera.sc

Ghoulboy78:
#341 raises the question of why this app is even here, given it's already built into carpet anyways.


altrisi:
It was decided/discussed between #187 and #188, though it hasn't really been updated since.

Ghoulboy78:
That's my point. Now I know that as the person who merged 188, I should probably have read through it properly first, but imo it made more sense to delete it.
cos if not, there is confusion for people who see two `camera.sc` apps, one built-in and one in the scarpet repo.

altrisi (approved):

rv3r (approved):
I agree with the reasoning that has been put forth.

PR #345, opened 2022-10-23
Move `CONTRIBUTING` and `README` to `.github` folder

altrisi:
Github still detects these there, and it cleans the root directory a bit, so users who don't know about github can find what matters quicker.

I used `.github` instead of `docs` because `.github` already exists, so we save an entry.

Also removes the indentation in contributing as it was making it a code block.


Ghoulboy78 (approved):
Agreed

replaceitem (approved):
Makes sense 👍

PR #346, opened 2022-10-24
Add pretty command to locate_block.sc

Ghoulboy78:


Ghoulboy78:
Just realised the app contains major flaws in the histogram functionality

PR #352, opened 2023-01-29
Prevent `simply_harvest.sc` from using up blocks in either hand

### LINKED ISSUE #351: simply_harvest off-hand block deletetion
**Describe the bug**
Off hand blocks are deleted while harvesting with simply_harvest

**To Reproduce**
1 place blocks in your off hand
2 harvest fully grown crops

**Expected behavior**
blocks in the off hand are not affected, moved to inventory or dropped if necessary. 

**Screenshots**
N/A

**Environment**
 - Carpet version: fabric-carpet-1.19.2-1.4.84+v221018
 - Minecraft version: 1.19.2

rv3r:
Fix #351

### DIFF
--- a/programs/survival/simply_harvest.sc
+++ b/programs/survival/simply_harvest.sc
@@ -16,6 +16,9 @@ global_seeds = {
     'nether_wart' -> 'nether_wart'
 };
 
+// variable indicating that harvesting occurred
+global_harvesting = false;
+
 __on_player_right_clicks_block(player, item_tuple, hand, block, face, hitvec) -> (
     if(hand=='offhand' || player~'gamemode_id' == 3, return());
     if(block_tags(block,'crops') || block == 'nether_wart',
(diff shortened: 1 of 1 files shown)


altrisi (approved):
Tested and fixes the issue! Don't know if it may be fragile in cases where multiple things happen per tick, but I think it's fine as that's unlikely anyway.

programs/survival/simply_harvest.sc
```diff
@@ -16,6 +16,9 @@ global_seeds = {
     'nether_wart' -> 'nether_wart'
 };
 
+// variable indicating that harvesting occurred
+global_harvesting = false;
+
```
Ghoulboy78, line 21: Where in the code do we actually query this variable tho?

altrisi: In __on_player_placing_block (ran after the main event handler if offhand) and in the 0-scheduled task (where it's just set back to false).

Ghoulboy78: Mah bad, didn't read properly

PR #355, opened 2023-01-29
Fix deprecations in sneak_grow_overpowered

### LINKED ISSUE #354: Deprecation warnings in `sneak_grow_overpowered`
It uses the `__on_tick_nether` and `__on_tick_ender` events.

altrisi:
Fixes #354.

This PR is intended to be *Rebase and merged*, given it includes a (very) backwards compatible way in the first commit, and a more general but modern one in the second one.

I think the second one is a good idea given it allows usage in custom and modded dimensions.

It'd probably be better to make it use sneak events that are dimension local, but that's for later.

### DIFF
--- a/programs/survival/sneak_grow_overpowered.sc
+++ b/programs/survival/sneak_grow_overpowered.sc
@@ -5,9 +5,15 @@ __config() -> (
    )
 );
 
-__on_tick() -> _sneak_grow();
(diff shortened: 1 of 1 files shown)


rv3r (approved):
Still works in all vanilla dimensions, deprecation warnings are not present.

Ghoulboy78 (approved):

PR #356, opened 2023-01-29
Add `README` to `utilities` section

altrisi:
Adds a `README` to the `utilities` folder.

Feel free to directly push changes if you know more about an app! Especially as I don't know about many.

Split from #289 given that show has been stopped since creation.


Ghoulboy78 (approved):

rv3r (approved):
I'll approve for now just so we have something there.

PR #359, opened 2023-02-01
Fix deprecated usage in `item_holding_mobs_despawn`

### LINKED ISSUE #358: `item_holding_mobs_despawn_rule` uses deprecated variant of `entity_load_handler`
The non-deprecated version takes another argument in the callback

altrisi:
Fixes #358.

Also removes the `rule` suffix in the name to prevent confusion with it being a Carpet rule, though let me know if I should revert that. It has no `README` entry so it doesn't need to be updated (adding missing entries to readme can be fixed later).

### DIFF
--- a/programs/survival/item_holding_mobs_despawn.sc
+++ b/programs/survival/item_holding_mobs_despawn.sc
@@ -8,7 +8,7 @@ __config() -> {
 };
 
 entity_load_handler('monster', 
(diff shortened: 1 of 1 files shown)


Ghoulboy78 (approved):

rv3r (approved):
Deprecation warning is gone, script still functions as expected.

PR #362, opened 2023-02-04
Prevent duplicate trades in `potion_master_cleric.sc`

### LINKED ISSUE #325: `potion_master_cleric.sc` can sometimes give the same trade in the same villager, and gives too few xp
By testing the script I've found it gave me the same trade (Potion of Grace) twice in the same villager and stage, with different prices.
A good enhancement could be for it to not be able to dupe the trade, given you'll obviously go for the cheaper one anyway.

I also think the potion trades give too little xp, according to the level and that they're not the simplest trades (by requiring an extra item).

rv3r:
Fixes #325 duplicate trade issue. I haven't touched xp produced by the trades.

Never get two trades of the same potion effect and potion type Maximum trades will never be more than maximum possible unique trades

### DIFF
--- a/programs/survival/potion_master_cleric.sc
+++ b/programs/survival/potion_master_cleric.sc
@@ -23,7 +23,7 @@ global_max_potion_trades = 3;
 // this code does not handle cycling villager trades. 
 global_potion_master_level = 3;
 
-// Nbt data for each avalible trade.
+// Nbt data for each available trade.
 global_potion_trades = [
 	// Dolphins Grace
 	'{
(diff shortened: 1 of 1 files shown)


programs/survival/potion_master_cleric.sc
```diff
@@ -130,15 +130,60 @@ global_potion_trades = [
...
+global_max_potion_trades = min(global_max_potion_trades,length(global_potion_trades) * length(global_potion_types));
+
 _add_trades(cleric, p, nbt) -> (
+	trades = {};
+	effects = copy(global_potion_trades);
 	// add 0-max number of potion trades
 	loop(rand(1 + global_max_potion_trades),
-
+		i = _;
+		mapped_effects = map(effects, _ ~ '(?<=Potion of )\\w*');
```
rv3r, line 142: > Why did you take the time to map the effects instead of using them directly?

Because debugging with a map that has keys that are 12+ lines of text is painful.

PR #364, opened 2023-02-04
"fewer", not "less"

rv3r:
I swear English is my first language.

### DIFF
--- a/programs/survival/portalorient.sc
+++ b/programs/survival/portalorient.sc
@@ -87,7 +87,7 @@ __on_player_changes_dimension(player, from_pos, from_dimension, to_pos, to_dimen
 
 		//scan each side of the portal for blocks based on the player's chosen mode
 		//	air -> face the side with more air blocks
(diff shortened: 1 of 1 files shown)


Ghoulboy78 (approved):
lol

altrisi:
Not English, would've used less too, is there a difference between them? In my head they mean the same.

Ghoulboy78:
Technically 'less' cannot be used as a comparative adjective, which is the context of the sentence. The sentence that rv3r is changing is referring to the side of the portal which has fewer solid blocks, while using 'less solid blocks' could imply that those blocks are less solid, instead of there being a smaller number of solid blocks.
In practice, the only people to care are English teachers, because as you mentioned, it makes sense to anyone who is reading the entire sentence.

But viva el pedanticism!

rv3r:
It's my script and the language is inconsistent, which is mostly why I fixed it.

altrisi (approved):

PR #366, opened 2023-02-12
Use air block above traced block to draw and color light level numbers

rv3r:
Line 92: shift the position to analyze up one block to...
Line 96: ...print the label on top of the block, not in the block
Line 97: ...get the light on top of the block, not in the block
Line 100: ...set the text color, again, based on the light on top of the block, not in the block

Upon further review, I might just dm suggestions since I don't know what I thought making a pull would do except, well, make one in the scarpet repo. Whoops!


Ghoulboy78 (approved):
Nah, this is better cos I can see wach individual change

PR #370, opened 2023-04-06
Add library for creating GUI menus

### LINKED ISSUE #315: Scarpet library for creating GUI menus.
After making the waystone app #305 and reviewing the fake player gui app #310
I think it would be nice to have a generalize library for GUI menus.

I like the idea of a `page` function and `button` or `icon`, and `filler` utilities. 
I also like the idea of a `open_page` function/ button

I think it should be as data generated as reasonable.
I am imaging passing in a map or array of `slot -> button/icon` maybe with optional callback lambdas called when clicked.

Ghoulboy78:
Much like `classes.scl`, this will in theory allow players to create their own GUI libraries by declaring a list of functions in a map.
Unlike `classes.scl`, this might actually be useful, since it abstracts a lot more of the grittier process behind nice-to use functions and syntax. Also, this has more flexibility.

Fixes #315

Todo items:
- [ ] Add basic (static) buttons which can do stuff when clicked
- [ ] Add dynamic buttons (clicking the button changes the icon, edits inventory itself, etc)
- [ ] Add storage? (so items u put in can be saved?)

### DIFF
--- a/programs/fundamentals/gui_menu.sc
+++ b/programs/fundamentals/gui_menu.sc
@@ -5,22 +5,6 @@ __command() -> (
     call_gui_menu(gui, player());
 );
 
-test()->(
-    screen = create_screen(player(),'generic_9x6',format('db Button Screen'),_(screen, player, action, data) -> (
-        //print(str('Action:\nAction: %s\nData: %s',action,data));
-        if(action=='quick_move'&&data:'slot'>=54,
-            return('cancel')
-        );
-        if(action=='pickup',
-            print(str('Action:\nAction: %s\nData: %s',action,data)); //for testing
-        );
-        //print(data);
-        if(data:'slot'==10,
-            'cancel'//prevent tampering with slot 10
-        );
-    ));
-    inventory_set(screen, 10, 1, 'stone')
-);
 
 global_inventory_sizes={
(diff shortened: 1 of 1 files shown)


opsaaaaa:
I just read the code. This looks pretty cool.
I've been wanting to rework the waystone app, when I do, Ill try this out.
Unfortunately I may not have time to try it out for awhile.
But I think it should tick the boxes.
Thanks.

opsaaaaa (approved):

Ghoulboy78:
https://github.com/gnembon/scarpet/pull/370/commits/55810ce4d50a7fb0570a91f4498d4d6fa17a7a71
Recording this commit as last commit before removal of placeholder names, JIC something goes wrong with that and I need to revert.

Ghoulboy78:
@altrisi I think this is ready

rv3r:
Seems handy and the test script demonstrates some cool functions. Thanks for adding the example else reviewing this would take an eternity. I'll being going through this for a while but I don't expect any major errors.

Ghoulboy78:
> Thanks for adding the example else reviewing this would take an eternity.

I actually created it so I could test the script as I added more and more to it. Imaging what a nightmare it would have been otherwise...

PR #371, opened 2023-04-15
Add notes to warning logs

Ghoulboy78:
Cos otherwise reading through logs can be a bit confusing if you don't know what you're looking for


rv3r (approved):
Looks good, ~but I'm not really in a position to test it. It'll probably be fine.~ It works and makes it easier to find these messages. Should this script prefix get added to line 40 as well?

rv3r (approved):
Still works

PR #373, opened 2023-04-27
Add AFK Display app

akaikagaribi:
Grays out a player's name if they're not moving for 3 minutes (configurable).
The functionality of this script is identical to that of Vanilla Tweaks' AFK Display datapack, but it saves what team a player was on before moving them to the AFK team, and puts them back in that team when they return.

We discovered that thing with AFK Display datapack when we started using teams, so I decided to write my own version in scarpet and publish it, because it may be useful for other people.

### DIFF
--- a/programs/survival/afk_display.sc
+++ b/programs/survival/afk_display.sc
@@ -1,0 +1,52 @@
+__config()->{
+    'scope'->'global',
+    'stay_loaded'->true
+};
+
+global_teams = read_file('teams.json', 'json');
+if(global_teams == null, global_teams = write_file('teams.json', 'json', {}); global_teams = read_file('teams.json', 'json'));
+
+// Change these values to your liking
+global_afk_prefix = '';
+global_afk_suffix = '';
+global_afk_timeout = 180; // seconds
+
+if(scoreboard('afkX') == null, scoreboard_add('afkX'));
+if(scoreboard('afkY') == null, scoreboard_add('afkY'));
+if(scoreboard('afkZ') == null, scoreboard_add('afkZ'));
+if(scoreboard('afkScore') == null, scoreboard_add('afkScore'));
+if(team_list('afk_players') == null,
+    team_add('afk_players');
+    team_property('afk_players', 'color', 'gray');
+    team_property('afk_players', 'prefix', global_afk_prefix);
+    team_property('afk_players', 'suffix', global_afk_suffix);
+);
+
+// To run this only each second
+global_tick = 0;
+__on_tick()->(
+    if((global_tick + 1) % 20 != 0, global_tick += 1; return(), global_tick = 0);
+    for(player('*'),
+        //run('say ' + global_teams:0);
+        if(and(
+                scoreboard('afkX', _) == scoreboard('afkX', _, _~'pos':0),
+                scoreboard('afkY', _) == scoreboard('afkY', _, _~'pos':1),
+                scoreboard('afkZ', _) == scoreboard('afkZ', _, _~'pos':2);
+            ),
+            scoreboard('afkScore', _, scoreboard('afkScore', _) + 1),
+            scoreboard('afkScore', _, 0);
+        );
+        if(scoreboard('afkScore', _) >= global_afk_timeout,
+            if(_~'team' == 'afk_players', continue());
+            global_teams:_ = _~'team';
+            write_file('teams.json', 'json', global_teams);
+            team_add('afk_players', _),
+
+            if(_~'team' != 'afk_players', continue());
+            if(global_teams:_, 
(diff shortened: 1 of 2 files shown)


programs/survival/afk_display.sc
```diff
@@ -0,0 +1,51 @@
...
+                scoreboard('afkY', _) == scoreboard('afkY', _, _~'pos':1),
+                scoreboard('afkZ', _) == scoreboard('afkZ', _, _~'pos':2);
+            ),
+            scoreboard('afkScore', _, scoreboard('afkScore', _) + 1),
+            scoreboard('afkScore', _, 0);
+        );
+        if(scoreboard('afkScore', _) >= global_afk_timeout,
+            if(_~'team' == 'afk_players', continue());
+            global_teams:_ = _~'team';
+            write_file('teams.json', 'json', global_teams);
```
Ghoulboy78, line 41: You can use __on_close() event to avoid writing to file each tick, which is rather slow and compute-expensive

programs/survival/afk_display.sc
```diff
@@ -0,0 +1,51 @@
...
+    if((global_tick + 1) % 20 != 0, global_tick += 1; return(), global_tick = 0);
+    for(player('*'),
+        if(and(
+                scoreboard('afkX', _) == scoreboard('afkX', _, _~'pos':0),
+                scoreboard('afkY', _) == scoreboard('afkY', _, _~'pos':1),
+                scoreboard('afkZ', _) == scoreboard('afkZ', _, _~'pos':2);
+            ),
+            scoreboard('afkScore', _, scoreboard('afkScore', _) + 1),
+            scoreboard('afkScore', _, 0);
+        );
```
Ghoulboy78, line 37: This works, but you could also use the `entity_event('on_move')` to detect when the player starts moving again. So basically, you have a map which is updated each time a player moves with the last time they move, and then uses `schedule()` to check if they still haven't moved after X ticks, and if so then mark them as afk.

Ghoulboy78:
This looks like a really cool app, which could possibly be moved to `useful` category instead of `survival`, but idk.

akaikagaribi:
Thank you for your feedback!

> You can use __on_close() event to avoid writing to file each tick, which is rather slow and compute-expensive

That's a good suggestion, I've done it out of concern that it won't let it save in case of crash, but it seems that it does.

> This works, but you could also use the `entity_event('on_move')` to detect when the player starts moving again. So basically, you have a map which is updated each time a player moves with the last time they move, and then uses `schedule()` to check if they still haven't moved after X ticks, and if so then mark them as afk.

This is also great, but despite trying to get better at scarpet by writing absurd scripts (with my main goal being to make poisonous potato the most OP item), I still can't get my head around the `entity_event()`. But this gave me an idea, so instead of  using `__on_tick()` event and skipping 19 calls out of 20, I rewrote it that `check_afk()` runs on start and it also `schedule()`s for 20 ticks after, which is clearly more performance friendly way of doing this.

About moving it to `useful` category, I believe you meant `utilites`, and I personally think it belongs more to `survival`, but I can move it anywhere else for sure if necessary.

programs/survival/afk_display.sc
```diff
@@ -0,0 +1,60 @@
...
+        if(and(
+            scoreboard('afkX', _) == scoreboard('afkX', _, _~'pos':0),
+            scoreboard('afkY', _) == scoreboard('afkY', _, _~'pos':1),
+            scoreboard('afkZ', _) == scoreboard('afkZ', _, _~'pos':2);
+            ),
+            scoreboard('afkScore', _, scoreboard('afkScore', _) + 1),
+            scoreboard('afkScore', _, 0);
+        );
+        if(scoreboard('afkScore', _) >= global_afk_timeout,
+            if(_~'team' == 'afk_players', continue());
```
rv3r, line 39: Does `continue()` even do anything in an `if`? From the wiki: 
```
break and continue can only be used inside for, c_for, while, loop, map, filter, reduce as well as Minecraft API block loops, scan and volume functions, while break can be used in first as well.```

altrisi: It's inside a `for` loop.

akaikagaribi: This fragment is inside the `for(player('*'), ...)` loop, so in this context it skips a player it currently iterates.

rv3r: I swear I triple checked for a loop. Lmao I'm blind

akaikagaribi:
I've found a way to get rid of any I/O operations, which is to return a player to their team whenever they leave, also I've added a few settings, such as whether to mark fake players, shadow players, and the color of the afk_players team.

PR #376, opened 2023-07-27
LockRotation app

TernaryC:
An app that recreates the behavior of the "lock rotation" tweak from the [Quark](https://quarkmod.net) mod.

This app adds a command, /lockrotation, which saves the player's current orientation for block placement (or an approximation), and applies it to all future blocks placed.
The lock can be disabled by running the command /lockrotation clear or setting the same rotation twice in a row.
Your current saved rotation can be displayed by running the command /lockrotation show.

I am not positive how this app will interact with any modded blocks, although I imagine not well. That's something that can be improved perhaps.
For vanilla blocks, though, I've accounted for as many edge cases as I could think of. Hopefully, I've not missed anything glaringly obvious.

### DIFF
--- a/programs/utilities/lockRotation.sc
+++ b/programs/utilities/lockRotation.sc
@@ -1,0 +1,207 @@
+///////////////////////////////////////////////////////////////////////////////
+// Rotation Lock
+// v1.0.0
+// scarpet script written by "TernaryC"
+// This description last updated 07/27/2023
+// 
+// This script aims to recreate the "lock rotation" feature of the Forge mod
+//  "Quark", as implemented in all versions of the mod since 1.14.4
+//
+// This script uses no source code from the "Quark" mod, nor does is it a port
+//   of that mod or it's modules.
+//
+// By running the command "/lockrotation", the player records their current
+//   facing direction, as well as other data related to block placement.
+// From then on, upon placing a block, the script will attempt to reorient the
+//   block to face the recorded placement data.
+//
+// The recorded data can be removed by typing the command "/lockrotation clear",
+//   or by locking into an identical rotation to the one already recorded (such
+//   as running the command twice in a row).
+//
+// The command "/lockrotation show" can be used to display the current
+//   recorded rotation data.
+//
+///////////////////////////////////////////////////////////////////////////////
+
+__on_start() ->
+(
+    global_lockProfile = null;
+);
+__command() -> _requestLock();
+
+// Front Facing
+show() ->
+(
+    details = _getDetails();
+    print(format('ig ' + details));
+    return();
+);
+_requestLock() ->
+(
+    _getCurrentState();
+    if (global_lockProfile == null || global_lockProfile:0 == null,
+        _tooltip(format(' Rotation lock ', 'b disabled'));
+    ,
+        _tooltip(format(' Rotation lock ', 'b enabled'));
+    );
+    return();
+);
+_getDetails() ->
+(
+    if (global_lockProfile == null,
+        return('[No lock set]')
+    );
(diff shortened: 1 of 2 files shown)


programs/utilities/lockRotation.sc
```diff
@@ -0,0 +1,238 @@
...
+    direction = global_lockProfile:0;
+    half      = global_lockProfile:1;
+    flat_dir  = global_lockProfile:2;
+    if (global_lockProfile == null, return(state));
+    
+    if (_contains(state, 'axis'), state:'axis' = _dirAxis(direction));
+    if (_contains(state, 'type'),
+        //Slab Type
+        if (_contains(['top', 'bottom'], state:'type') && half != -1,
+            state:'type' = halfSide(half);
```
rv3r, line 203: Function name should be `_halfSide` not `halfSide`

TernaryC: Shoot, I thought I caught all of those. Fixed it

rv3r:
Shaped stairs aren't getting updated when they get their rotation changed by this script. Locking rotation and trying to make shaped stairs should just produce nicely aligned stairs. Instead I get the attached image. Have you encountered this or tried to make stairs update properly?
stairs

rv3r:
Other than that it certainly functions as expected. Your `_castSide` function produces some fun results when `_sign` is passed 0 (especially since gnembon/fabric-carpet#1776 is a thing) but it still works so nbd.

I was a bit confused on how to get the "half" part of the lock to activate and I had to read through the code to find it, so an explanation of how to use the script could be added to the readme entry.

My last major note is that there are a lot of lines that could be optimized with scarpet-specific features (lines 109-117 could become `truepos = map(raycast - blockpos - 0.5,ceil(abs(_) - 0.49999) * _sign(_));`, ifs can act as their own elif, you can use function returns directly rather than assigning to a variable first, etc.), but it's already so fast that I'm not worried about these minor performance gains.

TernaryC:
I've added a short description to the readme, as well as a holographic indicator when running the command, which should make it a little more obvious how to use it.
I hadn't encountered the bug with shaped stairs, and while I can't seem to recreate your exact screenshot, I have found similar inconsistencies with the way stairs place when rotated.
The problem is easily resolved by simply delivering a block update to the glitchy stair block, but, of course, it doesn't seem to work if I update it programmatically. I might be missing something about the `update()` function?
I'll continue looking for a working solution that doesn't involve recreating the entire vanilla stair-shaping logic from scratch.

TernaryC:
Alright, I have something working. It's not ideal, but it does work and doesn't seem to have any adverse effects apart from an errant sfx.
I'm sure there must be a better way to do this, but I'll take a jank solution over no solution any day.

programs/utilities/lockRotation.sc
```diff
@@ -0,0 +1,347 @@
...
+		//	print(pos(_));
+		//	update(pos(_));
+		//);
+		//update(pos(global_rotated));
+		upos = [pos(global_rotated):0 + 1, pos(global_rotated):1, pos(global_rotated):2];
+		ublock = block(upos);
+		set(upos, 'air');
+		place_item('stone', upos);
+		set(upos, ublock);
+		global_rotated = null;
```
Ghoulboy78, line 345: Seen as though line 329 is the only place where the `global_rotated` variable becomes not null, can't you just move the code from `on_tick` into the if statement?
Or if you need a delay, use `schedule()`?

TernaryC: That snippet of code went through a lot of trial and error, and I basically stopped touching it once it worked. You're correct though, after testing it appears that code can, in fact, be moved to the if statement. I've made that change now.

PR #380, opened 2024-01-12
pos update / tp interference workaround

### LINKED ISSUE #377: cam.sc does not load pre-/cam location.
**Describe the bug**
When entering /cam all seems normal, however when you leave /cam it just changes your gamemode. see video: https://youtu.be/a6AkfL-PmV0

Video was taken with a fresh install of the /cam script.

Carpet settings: 
image

Mods running on the server: 
image

petersv5:
In 1.20.4 carpet it seems that a /tp command will override any entity pos update done in the same execution regardless of the order of the tp and pos update.

As a workaround update the position of the player in the world switching /tp instead of first switching worlds with a /tp and then position by a pos update.

This fixes #377

### DIFF
--- a/programs/survival/cam.sc
+++ b/programs/survival/cam.sc
@@ -117,9 +117,10 @@ __store_player_takeoff_params(player) ->
 
 __restore_player_params(player, config) ->
 (
-   run('execute in minecraft:'+config:'dimension'+' run tp @s ~ ~ ~');
+   [x, y, z] = config:'pos';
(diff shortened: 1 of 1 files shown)


Ghoulboy78 (approved):
Do you think this might affect keepalive.sc?

altrisi (approved):
Works, thanks!

Softchair:
Testing in 1.20.6. With the newest carpet 1.4.141 with the option "OpNoCheat" set to true the teleport command will not work as expected

PR #383, opened 2024-02-06
Fix shulkerboxes.sc for minecraft 1.20.4

MeeniMc:
the result of `parse_nbt` cannot be accessed with `:'text'` anymore

I tested with the following versions, and the change to the app appears backward compatible:
| Version | without PR | with PR |
|---------|--------------|---------|
| 1.20.4+carpet/1.4.128 |❌(script has no effect)|✅|
| 1.19.4+carpet/1.4.101 |✅|✅|
| 1.16.5+carpet/1.4.44 |✅|✅| 
| 1.15.2+carpet/1.4.8 |❌(script has syntax error)|❔|

### DIFF
--- a/programs/survival/shulkerboxes.sc
+++ b/programs/survival/shulkerboxes.sc
@@ -135,7 +135,7 @@ __add_item_to_vacuum_sboxes(player, search_item, refill_count, search_tag, do_ch
             if ( shulker_item ~ 'shulker_box$' 
                   && scount == 1 
                   && (nametag = shulker_tag:'display.Name') != null
-                  && lower(parse_nbt(nametag):'text') ~ 'vacuum'
+                  && lower(parse_nbt(nametag)) ~ 'vacuum'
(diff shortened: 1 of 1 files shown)


Ghoulboy78:
I don't think support for 1.15.2 is that big a deal rn, as long as it works in versions above it.

rv3r (commented):
Works in 1.16.5, 1.19.2, and 1.20.4 as stated. Only works after 1.4.18 if that matters

Ghoulboy78:
There seem to have been changes around that time with regards to how nbt values were read, which might explain it

MeeniMc:
I've been using this for a couple of weeks (on 1.20.4 server) and there has been no issues.

uolamer:
This is working for me on 1.20.4 server. Before I saw the PR I ended up coming up with the same fix through a bit of trial and error.

rv3r (approved):
Works as intended and is backwards compatible. Only works after carpet 1.4.18, which is fine by me.

PR #385, opened 2024-03-23
Added creative mode error message to cam.sc

### LINKED ISSUE #318: `cam` should give feedback when not entering because of invalid gamemode
For example, if you're in creative `cam` doesn't take you to camera mode, but it also doesn't give you any feedback at all, which can be confusing.

I think it should either complain about wrong gamemode or, like the old java implementation, switch to spectator anyway.

rv3r:
Fixes #318 by displaying an error message if you try to use the command while in creative mode. The message might not be perfect, feel free to comment on it
creative

### DIFF
--- a/programs/survival/cam.sc
+++ b/programs/survival/cam.sc
@@ -59,6 +59,8 @@ __command() ->
             display_title(p, 'actionbar', format('y Entered camera mode'));
          )
       ));
+   , current_gamemode == 'creative',
+      display_title(p, 'actionbar', format('r You must be in survival mode to use camera mode'));
    );
    null
 );


owensil (approved):

programs/survival/cam.sc
```diff
@@ -59,6 +59,8 @@ __command() ->
             display_title(p, 'actionbar', format('y Entered camera mode'));
          )
       ));
+   , current_gamemode == 'creative',
```
Ghoulboy78, line 62: At this point you might as well not add this, and just leave it as an end 'else' case, since even in adventure mode won't want to allow to use /cam

Ghoulboy78 (approved):

PR #387, opened 2024-05-21
Update potion master

owensil:
- Fixes NBT changes since file was last updated
- Spelling fixes and formatting


owensil:
Guess the NBT is different in 1.20.6 (was using 1.20.4) so going to close

PR #397, opened 2024-07-19
Create slimechunk.sc utility script

Maple38:
This script will tell the player whether they are in a slime chunk upon running the `/slimechunk` command. I find this to be a godsend on servers where you don't want players knowing the seed.

Don't really know how to code scripts, and I've never uploaded anything to GitHub before, please let me know if I'm doing something wrong :)

### DIFF
--- a/programs/utilities/slimechunk.sc
+++ b/programs/utilities/slimechunk.sc
@@ -1,0 +1,8 @@
+//Run the slimechunk to tell whether you are in a slime chunk
+__command() ->
+(
+    position = query(player(), 'pos');
+    is_slimechunk = in_slime_chunk(position);
+    if (is_slimechunk == 'true', exit('§aSlime chunk!'));
+    if (is_slimechunk == 'false', exit('§cNot a slime chunk :\('));
+);


programs/utilities/slimechunk.sc
```diff
@@ -0,0 +1,8 @@
+//Run the slimechunk to tell whether you are in a slime chunk
+__command() ->
+(
+    position = query(player(), 'pos');
+    is_slimechunk = in_slime_chunk(position);
+    if (is_slimechunk == 'true', exit('§aSlime chunk!'));
+    if (is_slimechunk == 'false', exit('§cNot a slime chunk :\('));
```
Ghoulboy78, line 7:
```suggestion
    if (is_slimechunk, '§aSlime chunk!', '§cNot a slime chunk :\(');
```

The `in_slime_chunk()` function doesn't return a string of `'true'` or `'false'`, rather it returns a boolean value (that is, the value `true` or `false`.
That means that you can just use the if/else statement to return a message based on whether or not it's a slime chunk.

Also, fyi, the `exit()` function is typically used in case of an error which means you can't continue with execution, not at the end of the function.

gnembon (approved):

PR #398, opened 2024-07-29
Fix shulkerboxes.sc 1.21 data components compability

Senth:
Due to changes in data components, the shulker boxes script stopped working in 1.20.5/1.21. This PR makes shulkerboxes.sc compatible with 1.21 (and possibly 1.20.5, but I haven't tested it)

### What changed
- Data components migration
  - `Count` is now lowercase `count`
  - Shulker inventory has been moved. `BlockEntityTag.Items` is now `components."minecraft:container"`
  - Custom data should be stored in `components."minecraft:custom_data"`
- Simplified some lines
- Consistency: Renamed variables to be more consistent across functions
- Formatting (remove trailing spaces, tabs were replaced with two spaces to be consistent with the rest of the file)

### Testing (only tested in 1.21)
- [x] The following modes that were affected
  - [x] vacuum
    - 1 item
    - full stack
    - full stack where shulker has would be filled up, remaining should be picked up by the player
  - [x] restock same
  - [x] restock keep
  - [x] restock keep vacuum
  - [x] swap next
  - [x] swap same
- [x] Potions refills
- [x] Totem of undying

### Note
I installed the vim plugin for scarpet, which reformatted the file, removing some trailing spaces. I can always split the PR into two if it needs to be more clear.

### DIFF
--- a/programs/survival/shulkerboxes.sc
+++ b/programs/survival/shulkerboxes.sc
@@ -18,14 +18,14 @@ will have vacuum capability
 Current vacuum mode:'+global_vacuum+'
 
 use /shulkerboxes toggle_vacuum to change it
-   
+
 Any shulkerbox with \'swap\' or \'restock\'
 followed with \'same\', \'keep\', \'first\',
 \'next\' or \'random\'
 will restock player inventory
  - swap: will swap a hotbar stack every time it changes
  - restock: will replace fully used up stacks
- 
+
  - same: will return only matching items
  - keep: as same but will keep 1 item in the shulkerbox
  - first: will always return first item from a box
@@ -35,7 +35,7 @@ will restock player inventory
 
 global_vacuum = 'collision';
 
-toggle_vacuum() -> 
+toggle_vacuum() ->
 (
    global_vacuum = if (global_vacuum == 'pickup', 'collision', 'pickup');
    'toggled vacuum mode to '+global_vacuum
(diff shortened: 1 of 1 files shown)


mal0andre (approved):
works perfectly

programs/survival/shulkerboxes.sc
```diff
@@ -83,11 +83,11 @@ __on_player_collides_with_entity(player, entity) -> if (global_vacuum == 'collis
 __item_animation(e, max_age, player) ->
 (
    particle('portal', pos(e)-[0,0.3,0], 10, 0.1, 0);
-   if (e~'age' > max_age, 
+   if (e~'age' > max_age,
       [item, count, tag] = e~'item';
       items_left = __add_item_to_vacuum_sboxes(player, item, count, tag, true);
       count_to_remove = count - items_left;
-      if (count_to_remove, modify(e, 'nbt_merge', '{Item:{Count:'+(count-count_to_remove)+'b}}'));
+      if (count_to_remove, modify(e, 'nbt_merge', '{Item:{count:'+(count-count_to_remove)+'}}'));
```
samule-i, line 90: Should Item be changed to item here?

Senth: I just remembered that I need to answer this one. `Item` should be uppercase. I tried with lowercase first but that did not work. My guess is that `{Item:` is part of scarpet and not Minecraft, but I'm not too familiar with either one.

lil-fox:
vacuum didn't work in pickup mode

lil-fox:
mb, it work, but only if you have full stack, insted of work even vith one item, like in original

gnembon (approved):

PR #401, opened 2024-08-13
autopickup.sh 2 item stack fix.

PooSmacker:
Added rv3r's fix in line 47, items will now stack to 64 instead of 2 before going into a new slow. All credits go to them.


WesleyJ-128:
This issue was solved by merging #412

Issue #408, opened 2024-09-16
keepalive.sc not working properly


FalseIokkon:
**Describe the bug**

**To Reproduce**
1. Spawn at least two players using /player <name> spawn at ~ ~ ~
2. Have keepalive.sc loaded
3. Stop and start the server
4. Only one player will be logged in.

Some issue when the server tries to respawn the first player, causing second and any following players to not respawn.
`[16:33:57] [Server thread/INFO]: Callback failed:          );
          HERE>> modify(player(_:'name'), 'flying', _:'fly')
      )
First argument to modify should be an entity in keepalive at line 17, pos 10`

**Expected behavior**
All players should properly respawn

**Environment**
 - Carpet version: fabric-carpet-1.21-1.4.147+v240623.jar
 - Minecraft version: 1.21

**Additional context**

botflakes:
Still bugged

tom-jm69:
Hey, whats the state on this?

tom-jm69:
Adding this to line 17 fixes the issue for me:

```sc         
         if (player(_:'name'),
            modify(player(_:'name'), 'flying', _:'fly')
         )
```

PR #410, opened 2024-09-23
fix: keepalive.sc - moved startup to a task

anaxios:
Moved startup to a task which allows players to load on server startup again.

https://github.com/gnembon/scarpet/issues/408


botflakes:
Even after applying your fixes, it still doesnt work properly

anaxios:
I'm using my version on my 1.20.1 server and it works. I'm not sure what else to do.

gnembon (approved):

PR #412, opened 2024-12-12
Fix auto_pickup.sc for new versions

WesleyJ-128:
At some point (potentially 1.20.6 when the component system was added), the nbt contained in item entities was changed to include the count of the items.  This breaks auto_pickup.sc, as it checks for the nbt of the items to match, which will not happen when the counts are unequal— my solution is to use ``replace()`` to remove the ``count`` field from the nbt before comparing.  This should be backwards-compatible as ``replace()`` will do nothing if the match is unsuccessful.


WesleyJ-128:
As the commit message notes, the regex fix deletes the counts in shulker box contents as well.  To fix this, I copy the nbt of the inventory item and modify its count to match that of the picked-up item, then compare if the modified nbt equals the picked-up item's nbt.

WesleyJ-128:
Fixed another issue with the ``stackableShulkerBoxes`` carpet rule— the pickup script would stack filled boxes together, which is incorrect.  There is now a change in behavior from previous versions; the script will automatically stack empty boxes, but this is convenient and empty boxes can be stacked in the inventory anyway.  If original functionality is desired, remove ``&& has(item_nbt, 'components')`` from line 45.

PR #413, opened 2024-12-13
Add an updated copy of refill_hotbar.sc

WesleyJ-128:
The existing refill_hotbar.sc is broken in versions 1.20.5+.  This is a fix added as a new script since it is not backwards-compatible.


gnembon:
I would suggest renaming the old one to _legacy, and keeping the always the new one working  as with normal name.

WesleyJ-128:
I've noticed what appears to be a client/server desync bug where the item swapped into the hotbar doesn't appear for the player if the item count is 1.  I don't know how/if this can be fixed.

gnembon:
Sorry, but now I am looking at it - the only difference really is in handling of the items of the sboxes. In that case this can be wrapped in a simple if based on major version number similar to 
https://github.com/gnembon/scarpet/commit/f848906e0e49ae8103399e8a8c7dbe4f2b80b35b
The scripts don't differ that much besides that nbt handling section
what do you think?

WesleyJ-128:
I like having just one script; I'll look at adding an if statement to combine the two.

PR #417, opened 2025-04-03
Updated the simply_harvest script

Gnottero:
Updated the simply harvest script in order to support all the blocks in the crops blocktag (plus nether warts), without hardcoding the max stage (using error handling with block states)


programs/survival/simply_harvest.sc
```diff
@@ -1,74 +1,77 @@
...
-__config() -> {'stay_loaded' -> true, 'scope' -> 'global'};
-
-global_seeds = {
-    'wheat' -> 'wheat_seeds',
-    'potatoes' -> 'potato',
-    'carrots' -> 'carrot',
-    'beetroots' -> 'beetroot_seeds',
-    'nether_wart' -> 'nether_wart'
+__config() -> {
+    'stay_loaded' -> true
```
gnembon, line 9: stay loaded is now implied, so no need to include non functional configs, right?

Gnottero: > stay loaded is now implied, so no need to include non functional configs, right?

Oh, I didn't know it was default now. I'll fix modify it later

Gnottero: I've updated the script removing the config block and also added another script, called simple_villagers, that allows the player to pickup villagers and place them back in the world

PR #421, opened 2025-05-01
Fix shulkerboxes.sc vacuum for items with nbt data

merrx:
Items with nbt data (e.g. renamed dirt) lose their data, if they get "vacuum"d.
This commit fixes this.


MeeniMc:
Thanks that fixes a very annoying bug indeed, especilly if you have a (separate) mod to make enchant  books stackable.

PR #426, opened 2025-05-17
added bots.sc

### LINKED ISSUE #388: keepalive.sc doesn't load fake players past the first one

### LINKED ISSUE #408: keepalive.sc not working properly

Rocka84:
Hi!
bots.sc is an enhanced replacement for keepalive.sc.
It can save and restore sets of bots (carpet players) and autosave/restore bots on server restart

It functionally replaces keepalive.sc therefore fixes #408 and fixes #388

### DIFF
--- a/programs/utilities/bots.sc
+++ b/programs/utilities/bots.sc
@@ -28,7 +28,7 @@ schedule(0, _() -> (
 ));
 
 _check_permission() -> (
-  if (player() == null, return(global_required_permission < 5));
+  if (player() == null || global_singleplayer, return(global_required_permission < 5));
   player()~'permission_level' >= global_required_permission;
 );
 
@@ -228,10 +228,12 @@ _on_exit() -> (
   ));
 );
 
-if (run('whitelist list'), ( //multiplayer
+if (system_info('server_ip') != null, (
+  global_singleplayer = false;
   __on_server_shuts_down() -> _on_exit();
   __on_server_starts() -> schedule(40, 'apply_set', '#autosave');
-), (                        //singleplayer
+), (
+  global_singleplayer = true;
   __on_close() -> _on_exit();
   __on_server_starts() -> schedule(1, 'apply_set', '#autosave');
 ));
--- a/programs/utilities/bots.sc
+++ b/programs/utilities/bots.sc
@@ -1,0 +1,293 @@
+// bots
+//  save and restore sets of bots (carpet players) and auto-restore bots on server restart
+// by Rocka84 (foospils)
+// based on keepalive.sc by gnembon
+// v1.4.1
+
+global_app_name = system_info('app_name');
+global_required_permission = 5;
+
+_print(msg) -> if(player(), print(player(), msg), print(msg));
+_print_formatted(...msg) -> _print(format(['bc '  + global_app_name + ': ', ...msg]));
+_error(msg) -> (
+  _print(format('br '  + global_app_name + ': ', 'r ' + msg));
+  exit();
+);
+
+_set_name(set) -> 'by 📂 ' + set;
+_bot_name(bot) -> 'bl 🤖 ' + bot;
+
+schedule(0, _() -> (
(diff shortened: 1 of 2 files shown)


programs/utilities/bots.sc
```diff
@@ -0,0 +1,293 @@
...
+);
+
+_on_exit() -> (
+  save_set('#autosave');
+  for (filter(player('all'), _~'player_type' == 'fake'), (
+    run('player ' + _~'name' + ' dismount');
+  ));
+);
+
+if (run('whitelist list'), ( //multiplayer
```
Rocka84: This is kind of hacky but works. The `whitelist` command only exists on servers so `run` returns `null` in singleplayer worlds.
Is there a better way to do this?

programs/utilities/bots.sc
```diff
@@ -0,0 +1,293 @@
...
+    'g' -> p~'gamemode',
+    'f' -> p~'flying',
+    'm' -> p~'mount'~'pos' || false,
+    'v' -> p~'mount'~'name' || false,
+    'r' -> _get_hand(p, 'mainhand'),
+    'l' -> _get_hand(p, 'offhand'),
+    's' -> p~'selected_slot',
+    'c' -> p~'sneaking',
+    'at' -> p~'swinging',
+    'u' -> false,
```
Rocka84, line 94: 'u' stands for 'using' and this is a planned feature. But I didn't find a way to determine it the player is continuously using an item yet.

programs/utilities/bots.sc
replaceitem: Maybe some options of `system_info` under the server category can be used in a similar way, where they only return `null` when not in multiplayer. `server_whitelist` also exists there.
https://github.com/gnembon/fabric-carpet/blob/master/docs/scarpet/Full.md#server-related-properties

Rocka84: Thanks for the tip! I now use `system_info('server_ip') != null`.
(Still a bit hacky but much better)

programs/utilities/bots.sc
huanqiugame: Hello, Rocka84! I have been experimenting with my own version of `keepalive.sc` or `logout_bot.sc`. I solve this problem by monitoring `__on_player_command(player, command)` event. Since controlling bots has to be done with commands, monitoring whether the player sends a command with `use continuous` or `use interval <tick>` is a workaround to the problem. However, this workaround fails when using extensions like Gugle Carpet Addition, or GCA for short, because it does not send a command when clicking a button for "use continous". The rest of this workaround contains monitoring player `stop` commands, `player_dies` event, and so on.

Your implementation is really inspiring to me, and [here is my implementation](https://github.com/huanqiugame/scarpet/blob/master/programs/utilities/logout_bot.sc) of logging bots out and in when players logging out and in. This script is pretty buggy but hopefully gives you some ideas.

